PR #8, opened 2023-08-20
Feat/rewrite FSRS item

L-M-Sherlock:


L-M-Sherlock:
@dae, would you mind I invite you to be a member of open spaced repetition? Some codes will needs your review.

dae:
Sure, no problem.

Your changes look fine, but you're gathering the items in two separate vecs and then creating a single vec - it would be simpler to store it in the single vec from the start. I can make a follow-up PR that will tidy it up a bit.

PR #9, opened 2023-08-20
Feat/weight clipper

### LINKED ISSUE #4: [TODO] feature: weight clipper
Python implementation:
https://github.com/open-spaced-repetition/fsrs-optimizer/blob/95694b787bb71ac9883db1201af09e334ee5ee0b/src/fsrs_optimizer/fsrs_optimizer.py#L100-L120

Luc-Mcgrady:
Not sure where to actually call the function but it passes the test I wrote. Hopefully that's what you wanted.
I'm not sure if the method of creating a new tensor is a good one but it seems to run fine.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -20,15 +20,12 @@ pub fn weight_clipper<B: Backend<FloatElem = f32>>(weights:Tensor<B, 1>) -> Tens
         (0.0, 1.0),
         (1.0, 10.0),
     ];
-    let mut i = 0; 
     // https://regex101.com/r/21mXNI/1
-    
 
     let val: &mut Vec<f32> = &mut weights.to_data().value;
 
-    for w in val.iter_mut().skip(4) {
+    for (i, w) in val.iter_mut().skip(4).enumerate() {
         *w = w.clamp(CLAMPS[i].0.into(), CLAMPS[i].1.into());
-        i += 1;
     } 
 
     Tensor::from_data(Data::new(val.clone(), weights.shape()))
--- a/src/model.rs
+++ b/src/model.rs
@@ -38,14 +38,14 @@ fn weight_clipper_test() {
 
     let tensor = Tensor::from_floats(
         [0.0, -1000.0, 1000.0, 0.0, // Ignored
-         1000.0, -1000.0, 1.0, 0.25]); // Clamped (1.0, 10.0),(0.1, 5.0),(0.1, 5.0),(0.0, 0.5),
+         1000.0, -1000.0, 1.0, 0.25, -0.1]); // Clamped (1.0, 10.0),(0.1, 5.0),(0.1, 5.0),(0.0, 0.5),
 
     let param: Tensor<Backend, 1> = weight_clipper(tensor);
     let values = &param.to_data().value;
 
     assert_eq!(*values, vec!
         [0.0, -1000.0, 1000.0, 0.0,
-         10.0, 0.1, 1.0, 0.25]);
+         10.0, 0.1, 1.0, 0.25, 0.0]);
 }
 
 #[derive(Module, Debug)]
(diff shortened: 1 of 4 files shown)


src/model.rs
```diff
@@ -93,7 +137,7 @@ impl<B: Backend<FloatElem = f32>> Model<B> {
             );
             let s_forget = self.stability_after_failure(stability, new_d.clone(), r);
             let new_s = s_recall.mask_where(rating.equal_elem(1), s_forget);
-            (new_s.clamp(0.1, 36500.0), new_d)
+            (new_s.clamp(0.1, 36500.0), weight_clipper(new_d))
```
L-M-Sherlock, line 140: It's not the position to call this function. Weight clipper should be applied to the parameters.

L-M-Sherlock: Could you call weight_clipper() in the end of forward()?

L-M-Sherlock (changes requested):

src/model.rs
Luc-Mcgrady: Like that?

src/model.rs
```diff
@@ -159,7 +159,7 @@ impl<B: Backend<FloatElem = f32>> Model<B> {
             // dbg!(&difficulty);
             // dbg!()
         }
-        (stability, difficulty)
+        (weight_clipper(stability), weight_clipper(difficulty))
```
L-M-Sherlock, line 162: The stability and difficulty are not the parameters. I mean `self.w`.

L-M-Sherlock (changes requested):

src/model.rs
L-M-Sherlock: `pub w: Param<Tensor<B, 1>>` is the parameter.

Luc-Mcgrady: Sorry. like that?

Luc-Mcgrady: hold on nvm

L-M-Sherlock:
Could you run this test?

https://github.com/open-spaced-repetition/fsrs-optimizer-burn/blob/19dc456680d950191f1c929c89eeed46a00e7314/src/training.rs#L132-L146

I get this error:

```
error[E0596]: cannot borrow `*self` as mutable, as it is behind a `&` reference
  --> src/training.rs:34:40
   |
34 |         let (stability, _difficulty) = self.forward(t_historys, r_historys);
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `self` is a `&` reference, so the data it refers to cannot be borrowed as mutable
```

Luc-Mcgrady:
Its because to change self.w I needed to make self mutable and that screwed up some code elsewhere.
Is there any way to change self.w without making self mutable?

L-M-Sherlock:
> Is there any way to change self.w without making self mutable?

I don't know. The maintainer said that they haven't supported it:

https://github.com/burn-rs/burn/issues/662#issuecomment-1684280084

Luc-Mcgrady:
I've given up and made it so that it just clips the weights at the very end.

https://github.com/Luc-Mcgrady/fsrs-optimizer-burn/blob/a6233d9b71ef0eeaacfb1481f716904afcba7593/src/training.rs#L118

I don't think replacing the tensor during training was a good idea anyway.

L-M-Sherlock:
> I don't think replacing the tensor during training was a good idea anyway.

But if we only clips the weights at the end, the errors could be larger than clipping the weights during training.

L-M-Sherlock (approved):

L-M-Sherlock:
I will merge this branch. It partially implemented the weight clipper.

Luc-Mcgrady:
I meant modifying the w values directly seems like a bad idea, I can see why you would want to clamp it in the forward function.

L-M-Sherlock:
But how could we modify the w values indirectly?

PR #12, opened 2023-08-21
Add clippy to CI checks

dae:

PR #13, opened 2023-08-21
Simplify Revlog->FSRSItem conversion

dae:
Depends on #12.

I wasn't entirely sure what the current code was trying to do, so please double-check it still performs as you expect / let me know if it's changed the behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,4 @@ log = "0.4"
 rusqlite = { version = "0.29.0" }
 chrono = "0.4.26"
 chrono-tz = "0.8.3"
+itertools = "0.11.0"
--- a/src/convertor.rs
+++ b/src/convertor.rs
@@ -193,12 +193,9 @@ fn convert_to_fsrs_items(
                     })
                     .collect();
                 FSRSItem {
-                    reviews,
-                    delta_t: entry.delta_t as f32,
-                    label: match entry.button_chosen {
-                        1 => 0.0,
-                        2 | 3 | 4 | _ => 1.0,
-                    },
+                    reviews: reviews,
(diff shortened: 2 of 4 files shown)


dae:
My earlier code was quite wrong, as I'd assumed there'd be only one FSRSItem for each card. The new code is passing the tests - does it look ok to you?

One remaining question - if we look at the shape of FSRSItem and FSRSReview, the delta_t and rating are now the same:

```rust
/// Represents a single review on a card, and contains the previous reviews for that card.
#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
pub struct FSRSItem {
    /// The previous reviews done to the card before the current one
    pub reviews: Vec<FSRSReview>,
    /// 1-4
    pub rating: i32,
    /// The number of days that passed
    pub delta_t: i32,
}

#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
pub struct FSRSReview {
    /// 1-4
    pub rating: i32,
    /// The number of days that passed
    pub delta_t: i32,
}
```

Would it make sense to remove the rating/delta_t from FSRSItem, and instead place one more review in the vec instead? Eg have it store current+previous reviews, and not just previous reviews.

L-M-Sherlock:
> Would it make sense to remove the rating/delta_t from FSRSItem, and instead place one more review in the vec instead?

Yeah, it's OK. I have read the code and found that you convert ratings to labels in FSRSBatcher. After insert the last rating/delta_t into vec, we need pop it in FSRSBatcher.

L-M-Sherlock (approved):
Nice work! I have reviewed the code and trained FSRS with the test collection.

PR #14, opened 2023-08-21
rewrite test for convertor

L-M-Sherlock:
@dae, I upload the collection.anki21 for test. Now you can check #13  whether pass the tests.


L-M-Sherlock:
OK. But I need to check whether the convertor is consistent with the python version. If the collection is too small, I'm not sure whether we have considered all boundary cases. I will remove the collection at first. Then I plan to collect boundary cases here. And create them in code in the future.

dae (approved):

PR #15, opened 2023-08-21
simplify

asukaminato0721:
I checked the implement, `get` has `clone`, so just use it.
use `.iter().next().expect("..")` is shorter than `for in + break`


L-M-Sherlock (approved):

L-M-Sherlock:
Please run cargo fmt before merging.

PR #18, opened 2023-08-22
Add BCELoss

L-M-Sherlock:


dae (approved):
I don't understand the math, but the Rust code looks fine. In the long run we may want to remove the dbg/info printing, but that's not urgent.

PR #23, opened 2023-08-22
add explanation for the sql query

asukaminato0721:
I tried to rewrite the sql query by params, but meet this problem.

```
---- convertor::test stdout ----
thread 'convertor::test' panicked at 'read error: SqlInputError { error: Error { code: Unknown, extended_code: 1 }, msg: "near \"?3\": syntax error", sql: "SELECT id, cid, ease, factor, type\n        FROM revlog \n        WHERE (type != 4 OR ivl <= 0)\n        AND id < ?1\n        AND cid < ?2\n        AND cid IN (\n            SELECT id\n            FROM cards\n            WHERE queue != 0\n            ?3\n            ?4\n        )", offset: 242 }', src/convertor.rs:230:37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- dataset::test_from_anki stdout ----
thread 'dataset::test_from_anki' panicked at 'read error: SqlInputError { error: Error { code: Unknown, extended_code: 1 }, msg: "near \"?3\": syntax error", sql: "SELECT id, cid, ease, factor, type\n        FROM revlog \n        WHERE (type != 4 OR ivl <= 0)\n        AND id < ?1\n        AND cid < ?2\n        AND cid IN (\n            SELECT id\n            FROM cards\n            WHERE queue != 0\n            ?3\n            ?4\n        )", offset: 242 }', src/convertor.rs:216:37

---- training::test stdout ----
thread 'training::test' panicked at 'read error: SqlInputError { error: Error { code: Unknown, extended_code: 1 }, msg: "near \"?3\": syntax error", sql: "SELECT id, cid, ease, factor, type\n        FROM revlog \n        WHERE (type != 4 OR ivl <= 0)\n        AND id < ?1\n        AND cid < ?2\n        AND cid IN (\n            SELECT id\n            FROM cards\n            WHERE queue != 0\n            ?3\n            ?4\n        )", offset: 242 }', src/convertor.rs:216:37
```

### DIFF
--- a/src/convertor.rs
+++ b/src/convertor.rs
@@ -48,12 +48,13 @@ fn read_collection() -> Result<Vec<RevlogEntry>> {
     };
 
     let current_timestamp = Utc::now().timestamp() * 1000;
-    // the query is safe since it will not input by user. The ?1 and ?2 is also unnecessary.
+    // the query is safe since it will not input by user. The ?1 and ?2 is also unnecessary. But to make others feel satefy.
     let revlogs = db
         .prepare_cached(&format!(
-            "SELECT id, cid, ease, factor, type
+            "SELECT id, cid, ease, type
         FROM revlog 
         WHERE (type != 4 OR ivl <= 0)
+        AND (factor != 0 or type != 3)
         AND id < ?1
         AND cid < ?2
         AND cid IN (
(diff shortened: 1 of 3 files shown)


dae:
? can only be used to replace single values, so ?3 and ?4 are invalid because they are replacing SQL. I would not worry too much about the SQL side of things for now - when this gets integrated into Anki, the revlog->FSRSItem conversion will happen on Anki's end, so this will be Anki's responsibility to handle.

dae:
(also note there are no security issues here, since the input SQL/values are either constructed by us, or constrained to numbers)

dae (approved):

asukaminato0721:
seems this is unnecessary, closed.

PR #29, opened 2023-08-25
Freezes weights

### LINKED ISSUE #20: [TODO] feature: freeze weights
Python implementation:

https://github.com/open-spaced-repetition/fsrs-optimizer/blob/95694b787bb71ac9883db1201af09e334ee5ee0b/src/fsrs_optimizer/fsrs_optimizer.py#L240-L244

Some weights could be calculated by other methods (not gradient descent). We should freeze them in the training process.

Reference: 
- https://discuss.pytorch.org/t/how-do-i-freeze-the-specific-weights-in-a-layer/104722/2
- https://discuss.pytorch.org/t/how-the-pytorch-freeze-network-in-some-layers-only-the-rest-of-the-training/7088/2

GraesonB:
closes #20 
I didn't write a specific test for this, but if you run it for one epoch you'll see that the first four weights don't change from what they are initialized to.

Also, Nathaniel mentioned that `grad_remove` should not be necessary here, but at the moment it's the only way to get `grad_replace` to work. He said he'll be writing tests tomorrow for it. 

I'm wondering if this should be configurable? As in whether the weights are frozen or not? Which weights are/aren't frozen?


L-M-Sherlock:
> I'm also wondering if this should be configurable? As in whether the weights are frozen or not? Which weights are/aren't frozen?

Only the first four weights may be frozen. I haven't implemented the pre-train module, where the first four weights will be trained.

L-M-Sherlock:
> Also, Nathaniel mentioned that `grad_remove` should not be necessary here, but at the moment it's the only way to get `grad_replace` to work. He said he'll be writing tests tomorrow for it.

OK, I will wait for his test.

GraesonB:
https://github.com/burn-rs/burn/pull/695

L-M-Sherlock:
Nice work! I will review the code tomorrow.

L-M-Sherlock:
I wrap your code into freeze_initial_stability() function to make it more readable.

By the way, I don't know whether it is possible that we can save a config to turn on/off freeze_initial_stability. The current implementation hasn't been prepared for freeze_initial_stability.

GraesonB:
Sounds good. It's certainly possible to make it configurable through the ModelConfig struct :-). I was able to do it locally, but I didn't include it in the PR because I was unsure of what you were envisioning. Let me know if that's what you would like and I can add it to the PR

L-M-Sherlock:
Thanks! Please include it in the PR.

GraesonB:
Added. Basically, instead of calling `ModelConfig::new()`, you initialize it like `ModelConfig: { freeze_stability: true }`. Right now, it defaults to false when initializing with `ModelConfig::new()`. 

In Burn's examples, they add the configurable paramaters inside `new`, but in my opinion that is less readable. e.g it would be `ModelConfig::new(false)` for this particular case. What are your thoughts?

dae:
I prefer the struct form too, though would recommend you have ModelConfig derive Default and set the default value to whatever we're more likely to use, so we can just do ModelConfig::default().

L-M-Sherlock (approved):
LGTM

GraesonB:
> I prefer the struct form too, though would recommend you have ModelConfig derive Default and set the default value to whatever we're more likely to use, so we can just do ModelConfig::default().

Burn seems to like setting defaults with their Config derive macro. So when calling `ModelConfig::new()`, it's using the defaults set with `#[config(default = false)]`. Is this sufficient, or would you also like to be able to use default?

dae:
The nice thing about default is when you have multiple options, and can do eg ModelConfig { freeze_stability: true, ..Default::default) } to get the defaults for the rest of the options.

Is freezing the most common use case? If so it would probably make sense to make true the default instead. Note that the config(default=...) I think only applies when creating a config from a JSON file, so we'd need to manually derive Default on it to get the right values.

One other possible improvement if other options are anticipated in the future: storing ModelConfig inside Model, instead of passing its individual options in to Model::new().

GraesonB:
That makes sense. My understanding is that this feature is staying off until the pre-train stage implemented. Admittedly I'm pretty new to FSRS, so I'm not sure what the pre-train stage entails, but I set the default to false for that reason. It sounds like once that's implemented freezing would be the normal behaviour, so yeah, maybe I should switch it the default to true?
What would training initialization look like if ModelConfig was inside Model? Are we planning to configure from a json file? Sorry if these are silly questions, still getting aquainted with the project.

dae:
Will defer to @L-M-Sherlock on sensible defaults; for the latter questions:

- instead of Model::new(freeze_stability) it would be Model::new(config)
- re JSON, it may be useful when testing/experimenting, but probably not so much when using this in third-party crates

L-M-Sherlock:
The pre-train stage will be used to optimize the values of initial stability. When it is finished, we will pass the first four values to the initialization of model and set freeze_stability: true.

PR #30, opened 2023-08-25
use path join

asukaminato0721:


L-M-Sherlock:
I'm not sure whether it is the best practice of Rust. You can ask the review from @dae.

dae:
The rest of the changes here seem reasonable, and it actually wouldn't be a big deal if this were merged without waiting - once Burn gets updated, it would just be a case of removing the .to_str().unwrap() to get non-UTF8 support.

dae (approved):

PR #35, opened 2023-08-28
Feat/clip weights during training

### LINKED ISSUE #4: [TODO] feature: weight clipper

L-M-Sherlock:
Close #4


dae (approved):
I can't see a better way to do this

PR #38, opened 2023-08-29
Fix/remove revlogs after manual reset & use const

L-M-Sherlock:
AFAIK, in the old version of Anki, revlogs didn't record manual reset. The related code in this PR is used to remove those revlogs after manual reset. In this case, the revlog's review_kind after manual reset is learning and the revlog's review_kind before manual reset is review or relearning.

### DIFF
--- a/src/convertor.rs
+++ b/src/convertor.rs
@@ -11,12 +11,16 @@ pub(crate) mod tests {
 
     use crate::dataset::{FSRSItem, FSRSReview};
 
+    const LEARNING: usize = 1;
+    const REVIEW: usize = 2;
+    const RELEARNING: usize = 3;
+
     #[derive(Clone, Debug, Default)]
     pub(crate) struct RevlogEntry {
         id: i64,
         cid: i64,
         button_chosen: i32,
-        review_kind: i64,
+        review_kind: usize,
         delta_t: i32,
     }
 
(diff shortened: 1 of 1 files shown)


src/convertor.rs
```diff
@@ -46,18 +50,22 @@ pub(crate) mod tests {
         next_day_starts_at: i64,
         timezone: Tz,
     ) -> Option<Vec<FSRSItem>> {
-        // Find the index of the first RevlogEntry in the last continuous group where review_kind = 0
-        // 寻找最后一组连续 review_kind = 0 的第一个 RevlogEntry 的索引
-        let mut index_to_keep = 0;
+        // Increment review_kind of all entries by 1
+        // 将所有 review_kind + 1
+        entries.iter_mut().for_each(|entry| entry.review_kind += 1);
```
dae, line 55: What do you think about the following alternative approach?

a) removing this
b) reducing the constants above by 1 each
c) adding a comment on the constants that they're one less than in the Python code, because we don't add 1 to everything

src/convertor.rs
```diff
@@ -46,18 +50,22 @@ pub(crate) mod tests {
...
+
+        // Find the index of the first RevlogEntry in the last continuous group where review_kind = LEARNING
+        // 寻找最后一组连续 review_kind = LEARNING 的第一个 RevlogEntry 的索引
+        let mut index_to_keep = LEARNING;
```
dae, line 59: This is an index, it should probably be 0 instead?

src/convertor.rs
```diff
@@ -46,18 +50,22 @@ pub(crate) mod tests {
...
 
         while i > 0 {
             i -= 1;
-            if entries[i].review_kind == 0 {
+            if entries[i].review_kind == LEARNING {
                 index_to_keep = i;
-            } else if index_to_keep != 0 {
-                // Found a continuous group of review_kind = 0, exit the loop
-                // 找到了连续的 review_kind = 0 的组，退出循环
+            } else if index_to_keep != LEARNING {
```
dae, line 66: Same here

dae:
Minor point: I find the "invalid" label confusing, as that's often used for when a bug report is not actually correct (eg user errors). Maybe "bug" would be a better label?

L-M-Sherlock:
Oh, it's my habit of labeling. I use bug for cases where the program is crashed, and invalid for cases where the program works but not as expected.

I don't insist my habit if you have a better labeling methodology.

dae:
I'd suggest "crash" for the former, and "bug" for the latter :-) But I don't have strong feelings; just wanted to let you know the current wording is a bit surprising.

src/convertor.rs
L-M-Sherlock: It's OK because review_kind is used to filter out data only.

L-M-Sherlock:
Maybe I misunderstand these descriptions😅:

image

I will use bug for the current case.

dae:
That short description is ambiguous, but if you see the longer description of 'invalid' here, it makes more sense: https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels

dae (approved):

PR #42, opened 2023-08-30
Feat/pre-training for initial stability

L-M-Sherlock:


src/pre_training.rs
```diff
@@ -0,0 +1,337 @@
...
+        data.sort_by(|a, b| a.0.cmp(&b.0));
+
+        let delta_t: Vec<_> = data.iter().map(|&(dt, _, _)| dt as f32).collect();
+        let recall: Vec<_> = data.iter().map(|&(_, r, _)| r as f32).collect();
+        let count: Vec<_> = data.iter().map(|&(_, _, c)| c as f32).collect();
+
+        let mut inner_result = HashMap::new();
+        inner_result.insert("delta_t".to_string(), delta_t);
+        inner_result.insert("recall".to_string(), recall);
+        inner_result.insert("count".to_string(), count);
```
dae, line 57: Could this be moved into a separate struct, instead of using untyped keys?

Otherwise, nice improvement :-)

src/model.rs
```diff
@@ -249,7 +257,7 @@ mod tests {
     fn forward() {
         use burn_ndarray::NdArrayBackend;
         type Backend = NdArrayBackend<f32>;
-        let model = Model::<Backend>::new(false);
+        let model = Model::<Backend>::new(false, None);
```
dae, line 260: After this PR lands, I'll we refactor this so the config is passed in instead.

src/pre_training.rs
```diff
@@ -0,0 +1,131 @@
...
+    let weighted_diff = &diff * &diff * count;
+    let rmse = f32::sqrt(weighted_diff.sum() / count.sum());
+    // TODO: add l1 regularization
+    rmse
+}
+
+fn search_parameters(pretrainset: HashMap<i32, HashMap<String, Vec<f32>>>) -> HashMap<i32, f32> {
+    let mut optimal_stabilities: HashMap<i32, f32> = HashMap::new();
+
+    let epsilon = 1e-6; // precision
```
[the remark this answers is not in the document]
asukaminato0721: well, this will be into inf loop.

src/pre_training.rs
```diff
@@ -142,10 +142,7 @@ fn smooth_and_fill(
 
     let mut init_s0 = vec![];
 
-    let r_s0_default = [(1, 0.4), (2, 0.6), (3, 2.4), (4, 5.8)]
-        .iter()
-        .cloned()
-        .collect::<HashMap<_, _>>();
+    let r_s0_default = HashMap::from([(1, 0.4), (2, 0.6), (3, 2.4), (4, 5.8)]);
```
dae, line 145: Nice

src/pre_training.rs
```diff
@@ -188,49 +185,48 @@ fn smooth_and_fill(
...
+                        rating_stability[&2].powf(1.0 / w1)
+                            * rating_stability[&3].powf(1.0 - 1.0 / w1),
                     );
                 }
                 (true, false, false, true) => {
                     rating_stability.insert(
                         2,
-                        f32::powf(rating_stability[&1], w1 / (w1 + w2 - w1 * w2))
-                            * f32::powf(rating_stability[&4], 1.0 - w1 / (w1 + w2 - w1 * w2)),
+                        rating_stability[&1].powf(w1 / (w1 + w2 - w1 * w2))
```
dae, line 200: Nice

asukaminato0721 (approved):

asukaminato0721:
Some notes: the huge match is learned from https://github.com/python/cpython/blob/3.11/Lib/dataclasses.py#L863

src/pre_training.rs
dae: The rest looks good, but I still think this should use a Struct instead of a HashMap - is there any reason it cannot?

PR #49, opened 2023-09-01
Use an enum instead of a hashmap for table lookups

dae:


L-M-Sherlock (approved):
Nice, it's more readable and neat than before.

PR #50, opened 2023-09-01
more izip

asukaminato0721:
less noise.

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -116,27 +116,26 @@ fn simulate(config: &SimulatorConfig, request_retention: f64, seed: Option<u64>)
         let mut delta_t = Array1::zeros(deck_size); // Create an array of the same length for delta_t
 
         // Calculate delta_t for entries where has_learned is true
-        Zip::from(&mut delta_t)
-            .and(&old_last_date)
-            .and(&has_learned)
-            .for_each(|delta_t, &last_date, &has_learned_flag| {
-                if has_learned_flag {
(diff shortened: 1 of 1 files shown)


src/optimal_retention.rs
```diff
@@ -290,19 +284,21 @@ fn simulate(config: &SimulatorConfig, request_retention: f64, seed: Option<u64>)
...
             });
-
-        Zip::from(&mut new_difficulty)
-            .and(&ratings)
-            .and(&true_learn)
-            .for_each(|new_diff, &rating, &true_learn_flag| {
-                if true_learn_flag {
-                    *new_diff = w[4] - w[5] * (rating as f64 - 3.0);
-                }
-            });
```
dae, line 301: One downside of Rust macros is that they can prevent proper automatic formatting. I presume you've manually added the spaces so than "in" is aligned here? Cargo fmt will presumably not be able to fix these lines, so if things are renamed in the future, we'll need to manually format them again.

I do think your changes are an improvement from the long lines below, but you might want to consider just switching to Zip::from() everywhere instead, as that way everything will format automatically.

Just my 2c :-)

asukaminato0721: make senses, now I let >= 4 lines' `Zip` to use `azip!`. Since macro is bad at code formatting and code check/diagnose.

asukaminato0721: ohh, I find that `izip` is much better, it can be formatted.

L-M-Sherlock (approved):
LGTM

PR #58, opened 2023-09-08
Feat/support scheduler

L-M-Sherlock:
Three new API: calc_memo_state, next_interval, next_memo_state

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -27,7 +27,7 @@ fn infer<B: Backend<FloatElem = f32>>(
     (stability, difficulty, retention)
 }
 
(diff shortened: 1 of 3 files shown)


asukaminato0721 (approved):

dae (approved):
Looks good! I'll add a benchmark and make some minor tweaks in a follow-up - performance is very good.

dae:
Quick question: what is the 'i' argument in next_memo_state() for?

L-M-Sherlock:
It's the length of review history. It's should be zero for the first learning of a new card.

PR #59, opened 2023-09-09
Benchmark and API tweaks

dae:
What do you think about the new API?

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -15,6 +15,9 @@ use crate::model::Model;
 use crate::training::BCELoss;
 use crate::{FSRSError, FSRSItem, FSRSReview};
 
+/// This is a slice for efficiency, but should always be 17 in length.
+type Weights = [f32];
+
 fn infer<B: Backend<FloatElem = f32>>(
     model: &Model<B>,
     batch: FSRSBatch<B>,
@@ -27,7 +30,7 @@ fn infer<B: Backend<FloatElem = f32>>(
     (stability, difficulty, retention)
 }
 
-fn weights_to_modela(weights: &[f32; 17]) -> Model<NdArrayBackend<f32>> {
+fn weights_to_modela(weights: &Weights) -> Model<NdArrayBackend<f32>> {
     type Backend = NdArrayBackend<f32>;
     let config = ModelConfig::default();
     let mut model = Model::<Backend>::new(config);
@@ -38,7 +41,7 @@ fn weights_to_modela(weights: &[f32; 17]) -> Model<NdArrayBackend<f32>> {
     model
 }
 
-pub fn calc_memo_state(weights: &[f32; 17], item: FSRSItem) -> (f32, f32) {
+pub fn calc_memo_state(weights: &Weights, item: FSRSItem) -> (f32, f32) {
     type Backend = NdArrayBackend<f32>;
     let model = weights_to_modela(weights);
     let (time_history, rating_history) = item.reviews.iter().map(|r| (r.delta_t, r.rating)).unzip();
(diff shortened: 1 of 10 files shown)


dae:
Hmm. Thinking more about it, I wonder whether using the saved stability might be better than passing in the whole revlog. In https://github.com/open-spaced-repetition/rs-fsrs/issues/6#issuecomment-1711109835 you said that this would be incorrect when the user changes their weights. But when they change their weights, couldn't we automatically apply the updated stability to all cards using that preset? This would mean less work needs to be done during review, and would be easier to integrate into Anki, as Anki's state handling code doesn't currently have access to the revlog history. WDYT?

L-M-Sherlock:
> But when they change their weights, couldn't we automatically apply the updated stability to all cards using that preset?

Yes, we can use `calc_memo_state` here to update them in batch. It's more efficient and convenient.

dae:
Great, I'll refactor this PR.

dae:
One remaining problem is that without access to the revlog history, we don't know what `i` should be. And we can't just do a simple "select count(*) from revlog where cid=?", because we need the logic that discards reviews after rescheduling. It would be nice if we could avoid looking at the revlog during review.

Maybe one way to optimize this would be to store `i` as another custom_data  value, and remove it when manually rescheduling.  If missing from custom_data, it could be calculated based on the revlog history at review time. Do you foresee any problems with such an approach?

L-M-Sherlock:
> One remaining problem is that without access to the revlog history, we don't know what `i` should be.

Actually, it's not necessary to get the exact length of the revlog history. `i` is used to distinguish whether it is the first review. Because in the first review, previous memo state doesn't exist, and FSRS need to initialize the memo state from the first rating.

dae:
I've refactored things so there's a top-level Fsrs struct that users will import, and it can use like this:

```rust
        let fsrs = Fsrs::new(Some(WEIGHTS));

        # after weight change
        let state = fsrs.memory_state(item);
    
        # during review
        let states = fsrs.next_states(Some(state), 0.9, 21),
```

Does that work for you?

Also since this crate now does both inference and training, what do you think about renaming the repo to fsrs-rs?

L-M-Sherlock:
> what do you think about renaming the repo to fsrs-rs?

I will change it after merging.

> Does that work for you?

I'm reviewing it in my local device.

src/model.rs
```diff
@@ -303,3 +316,48 @@ mod tests {
         )
     }
 }
+
+/// This is the main structure provided by this crate. It can be used
+/// for both weight training, and for reviews.
+pub struct Fsrs<B: Backend<FloatElem = f32> = NdArrayBackend> {
```
L-M-Sherlock, line 322: It looks weird that only `F` is capital. FSRS is abbreviation of free spaced repetition scheduler. I recommend using `FSRS` here.

src/inference.rs
```diff
@@ -13,82 +10,176 @@ use crate::dataset::FSRSBatcher;
...
+            let true_val = batch.labels.clone().float().to_data().value;
+            all_true_val.extend(true_val);
+            all_retention.push(retention);
+            all_labels.push(batch.labels);
+            progress_info.current += chunk.len();
+            if !progress(progress_info) {
+                return Err(FSRSError::Interrupted);
+            }
+        }
+        let rmse = calibration_rmse(all_predictions, all_true_val);
```
L-M-Sherlock, line 148: To avoid confusion with the common rmse, I recommend renaming `rmse` to `rmse_bins`. And keep its consistency in Anki (RMSE -> RMSE(bins)).

L-M-Sherlock (changes requested):

L-M-Sherlock:
Besides, I meet this error in my vscode:

```
error[E0432]: unresolved import `fsrs_optimizer`
  --> benches/benchmark.rs:10:5
   |
10 | use fsrs_optimizer::FSRSReview;
   |     ^^^^^^^^^^^^^^ use of undeclared crate or module `fsrs_optimizer`
```

L-M-Sherlock (approved):
LGTM

PR #66, opened 2023-09-12
some improvements

asukaminato0721:
I noticed that `row_to_revlog_entry` is more like a `TryFrom`, so I did some change.

And also the `Path`.


dae (approved):
LGTM

L-M-Sherlock (approved):

Issue #1, opened 2023-08-17
[TODO] fix:arguments to clamp are incorrect


L-M-Sherlock:
https://github.com/open-spaced-repetition/fsrs-optimizer-burn/blob/4d4555ed7c00aa9a0b6e09c1c011ccf058ce3e7c/main.rs#L57

```
error[E0308]: arguments to this method are incorrect
   --> main.rs:57:31
    |
57  | ...   let new_d = new_d.clamp(1.0, 10.0); // TODO: consider constraining the associated type `<B as Backend>::FloatElem` to `{float}` or ...
    |                         ^^^^^
    |
note: expected associated type, found floating-point number
   --> main.rs:57:37
    |
57  | ...   let new_d = new_d.clamp(1.0, 10.0); // TODO: consider constraining the associated type `<B as Backend>::FloatElem` to `{float}` or ...
    |                               ^^^
    = note: expected associated type `<B as Backend>::FloatElem`
                          found type `{float}`
    = help: consider constraining the associated type `<B as Backend>::FloatElem` to `{float}` or calling a method that returns `<B as Backend>::FloatElem`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
note: expected associated type, found floating-point number
   --> main.rs:57:42
    |
57  | ...   let new_d = new_d.clamp(1.0, 10.0); // TODO: consider constraining the associated type `<B as Backend>::FloatElem` to `{float}` or ...
    |                                    ^^^^
    = note: expected associated type `<B as Backend>::FloatElem`
                          found type `{float}`
    = help: consider constraining the associated type `<B as Backend>::FloatElem` to `{float}` or calling a method that returns `<B as Backend>::FloatElem`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
note: method defined here
   --> /Users/jarrettye/.cargo/git/checkouts/burn-acfbee6a141c1b41/8808ee2/burn-tensor/src/tensor/api/numeric.rs:414:12
    |
414 |     pub fn clamp(self, min: K::Elem, max: K::Elem) -> Self {
    |            ^^^^^
```

L-M-Sherlock:
@dae, @asukaminato0721, could you help me?

dae:
[...]

PR #67, opened 2023-09-13
Drop constraint on FloatElem

dae:
https://github.com/open-spaced-repetition/fsrs-rs/issues/1#issuecomment-1681801153 has been addressed by Burn:
https://github.com/burn-rs/burn/issues/795

With this change, if we wanted to switch to f16/f64 in the future, we'd only need to change the backend creation.


L-M-Sherlock (approved):
LGTM

PR #68, opened 2023-09-13
add badge, add tests

asukaminato0721:
add some tests, and some small change.

### DIFF
--- a/src/batch_shuffle.rs
+++ b/src/batch_shuffle.rs
@@ -69,7 +69,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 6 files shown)


src/batch_shuffle.rs
```diff
@@ -22,15 +22,15 @@ where
 
         // Create a vector of batch indices and shuffle it
         // 创建一个批数索引的向量并打乱
-        let mut batch_indices: Vec<usize> = (0..num_batches).collect();
+        let mut batch_indices: Vec<_> = (0..num_batches).collect();
```
dae, line 25: One alternative is to use itertools' .collect_vec() instead

dae (approved):
LGTM

PR #69, opened 2023-09-13
Update burn

dae:
The first commit should fix the old dependency issues.

The second commit is for discussion, and I'm happy to change it if you prefer the old way.

-  do either of you currently use the experiment.log or checkpoint file outputs in the artifact dir? Or even the config.json? If not, we may gain a small performance boost by not writing those files - especially experiment.log, which can be quite big.
- the new train CLI is a bit distracting when running tests. What do you think about hiding progress when the tests are run, as I've done in the second commit?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,30 +56,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anstyle"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "15c4c2c83f81532e5845a733998b6971faca23490340a418e9b72a3ec9de12ea"
-
-[[package]]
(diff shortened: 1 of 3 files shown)


L-M-Sherlock:
> * do either of you currently use the experiment.log or checkpoint file outputs in the artifact dir?

I only use the experiment.log and checkpoint file for debugging.

dae:
Do you use it frequently enough that you'd like it on by default, or would you be happy to do something like 'cargo test --release --features log' when you need it?

dae:
or maybe LOG=1 cargo test --release

L-M-Sherlock:
Yeah, it's OK. By the way, I couldn't find the log file when I use Anki fsrs branch for test. Does it mean that it doesn't write anything?

dae:
Yeah, I turned it off for Anki, since we don't need it. But if you'd find it useful, I could make it configurable via an env var.

L-M-Sherlock:
It's better to make it configurable. `NaN` doesn't raise error in the model, so it's hard to locate the bug if we doesn't have the log file during the training.

dae:
Ok, I'll update this PR

dae:
Do you have a preference as to whether the text goes to stdout or a file? With an env var like:

`RUST_LOG=debug ./run`, you could see the text in real time on the terminal, or use `RUST_LOG=debug ./run | tee file` if you also wanted to save it to a file.

L-M-Sherlock:
> Do you have a preference as to whether the text goes to stdout or a file?

I prefer to output them in file, because the training process often generates tons of logs, which will flood the terminal.

dae:
Does that work for you? It should work in Anki too, once I update it.

L-M-Sherlock (approved):
LGTM.

Usage:

```rust
BURN_LOG=tmp/fsrs cargo test --release
```

PR #70, opened 2023-09-16
Avoid updating memory state when no days elapsed

dae:
My goal is to solve https://github.com/ankitects/anki/pull/2654#issuecomment-1718600639 and support interday learning steps as well. If we do it this way, we should be able to use next_states on every answer (both learning+review), and we shouldn't need to check the revlog when graduating. What do you think?


L-M-Sherlock (approved):
LGTM

PR #73, opened 2023-09-17
Feat/extract optimal retention parameters from revlog

L-M-Sherlock:


inRm3D (approved):

asukaminato0721 (approved):

src/convertor_tests.rs
```diff
@@ -246,6 +248,123 @@ fn read_collection() -> Result<Vec<RevlogEntry>> {
...
+        .counts_by(|r| r.button_chosen);
+    let first_rating_prob = first_rating_count
+        .iter()
+        .map(|(button_chosen, count)| {
+            (
+                *button_chosen,
+                *count as f32 / first_rating_count.values().sum::<usize>() as f32,
+            )
+        })
+        .collect::<HashMap<_, _>>();
```
dae, line 266: Because we know there are only 4 possible values here, it would be simpler to collect directly into an array here. You can use this helper which I'm happy to license as BSD-3 here: https://github.com/ankitects/anki/blob/59759b468f03da47f33ea2867c0a0f773e8fcd02/rslib/src/scheduler/fsrs/try_collect.rs#L11

Like this:

let first_rating_prob: [f32; 4] = ...try_collect().unwrap()

src/convertor_tests.rs
```diff
@@ -246,6 +248,123 @@ fn read_collection() -> Result<Vec<RevlogEntry>> {
...
+                *button_chosen,
+                *count as f32 / review_rating_count.values().sum::<usize>() as f32,
+            )
+        })
+        .collect::<HashMap<_, _>>();
+    let review_rating_prob = [
+        review_rating_prob.get(&2).copied().unwrap_or_default(),
+        review_rating_prob.get(&3).copied().unwrap_or_default(),
+        review_rating_prob.get(&4).copied().unwrap_or_default(),
+    ];
```
dae, line 291: Same here

src/convertor_tests.rs
```diff
@@ -246,6 +248,123 @@ fn read_collection() -> Result<Vec<RevlogEntry>> {
...
+        .copied()
+        .unwrap_or_default()
+        + recall_costs.get(&1).copied().unwrap_or_default();
+    assert_eq!(forget_cost, 23.481838);
+
+    let recall_costs = [
+        recall_costs.get(&2).copied().unwrap_or_default(),
+        recall_costs.get(&3).copied().unwrap_or_default(),
+        recall_costs.get(&4).copied().unwrap_or_default(),
+    ];
```
dae, line 364: And here

dae (commented):
Thanks for looking into this guys.

src/convertor_tests.rs
dae: invalid_input!() would need to be changed to return Err(FSRSError::InvalidInput)

src/convertor_tests.rs
```diff
@@ -255,55 +254,60 @@ fn extract_simulator_config_from_revlog() {
...
+                    *count as f32 / first_rating_count.values().sum::<usize>() as f32,
+                )
+            })
+            .for_each(|(i, v)| {
+                arr[i as usize] = v;
+            });
+        arr
+    };
+    assert_eq!(
+        first_rating_prob[1..=4],
```
dae, line 273: Might be simpler to .filter() out the button_chosen=0 case

dae: How about:

            .for_each(|(i, v)| {
                if i > 0 { arr[i-1 as usize] = v; }
            });

dae:
Thanks, looks good. What do you think about these minor tweaks?

```diff
diff --git a/src/convertor_tests.rs b/src/convertor_tests.rs
index c850480..1a4f6d4 100644
--- a/src/convertor_tests.rs
+++ b/src/convertor_tests.rs
@@ -250,16 +250,16 @@ fn extract_simulator_config_from_revlog() {
     let revlogs = read_collection().unwrap();
     let first_rating_count = revlogs
         .iter()
-        .filter(|r| r.review_kind == RevlogReviewKind::Learning && r.ease_factor == 0)
+        .filter(|r| r.review_kind == RevlogReviewKind::Learning && r.ease_factor == 0 && r.button_chosen > 0)
         .counts_by(|r| r.button_chosen);
+    let total_ratings = first_rating_count.values().sum::<usize>() as f32;
     let first_rating_prob = {
-        let mut arr = [Default::default(); 4];
+        let mut arr = [0.0; 4];
         first_rating_count
             .iter()
-            .filter(|(&button_chosen, ..)| button_chosen >= 1)
             .for_each(|(button_chosen, count)| {
                 arr[*button_chosen as usize - 1] =
-                    *count as f32 / first_rating_count.values().sum::<usize>() as f32
+                    *count as f32 / total_ratings
             });
         arr
     };
@@ -269,7 +269,7 @@ fn extract_simulator_config_from_revlog() {
         .filter(|r| r.review_kind == RevlogReviewKind::Review && r.button_chosen != 1)
         .counts_by(|r| r.button_chosen);
     let review_rating_prob = {
-        let mut arr = [Default::default(); 3];
+        let mut arr = [0.0; 3];
         review_rating_count
             .iter()
             .filter(|(&button_chosen, ..)| button_chosen >= 2)
@@ -282,10 +282,10 @@ fn extract_simulator_config_from_revlog() {
     assert_eq!(review_rating_prob, [0.07380187, 0.90085745, 0.025340684,]);
 
     let recall_costs = {
-        let mut arr = [Default::default(); 5];
+        let mut arr = [0.0; 4];
         revlogs
             .iter()
-            .filter(|r| r.review_kind == RevlogReviewKind::Review)
+            .filter(|r| r.review_kind == RevlogReviewKind::Review && r.button_chosen > 0)
             .sorted_by(|a, b| a.button_chosen.cmp(&b.button_chosen))
             .group_by(|r| r.button_chosen)
             .into_iter()
@@ -293,7 +293,7 @@ fn extract_simulator_config_from_revlog() {
                 let group_vec = group.into_iter().map(|r| r.taken_millis).collect_vec();
                 let average_secs =
                     group_vec.iter().sum::<u32>() as f32 / group_vec.len() as f32 / 1000.0;
-                arr[button_chosen as usize] = average_secs
+                arr[button_chosen as usize - 1] = average_secs
             });
         arr
     };
@@ -329,7 +329,7 @@ fn extract_simulator_config_from_revlog() {
         for (review_kind, sec) in review_kind_to_total_millis.into_iter() {
             group_sec_by_review_kind[review_kind as usize].push(sec)
         }
-        let mut arr = [Default::default(); 5];
+        let mut arr = [0.0; 5];
         group_sec_by_review_kind
             .iter()
             .enumerate()
@@ -340,10 +340,10 @@ fn extract_simulator_config_from_revlog() {
         arr
     };
 
-    let forget_cost = forget_cost[RevlogReviewKind::Relearning as usize] + recall_costs[1];
+    let forget_cost = forget_cost[RevlogReviewKind::Relearning as usize] + recall_costs[0];
     assert_eq!(forget_cost, 23.481838);
 
-    assert_eq!(recall_costs[2..=4], [9.047336, 7.774851, 5.149275,]);
+    assert_eq!(recall_costs[1..=3], [9.047336, 7.774851, 5.149275,]);
 }
 
 // This test currently expects the following .anki21 file to be placed in tests/data/:
```

L-M-Sherlock:
Is it ready to merge?

dae:
Yep - thanks again!

dae:
first_answer_prob[3] was always 0 in Anki, because when easy is used, ease_factor is set. In porting it to Anki, I've changed it to r.last_interval == 0 instead, which seems to work.

PR #74, opened 2023-09-18
Clip user-provided weights

dae:
Closes https://github.com/ankitects/anki/issues/2666


dae:
Doing it in this crate means that the values displayed inside Anki may be outside the clipping bounds, but it ensures that values are clipped even if the user directly modifies their .anki2 file using third-party software.

L-M-Sherlock (approved):

PR #82, opened 2023-09-26
memory_state_from_sm2() and next_interval()

dae:


L-M-Sherlock (approved):
LGTM. I added an extra test for ensure the FSRS factor is near the ease factor. @dae, are you ready to merge it?

dae:
Yep! If it’s not marked as a draft, I consider it good to go

user1823:
As per the description in https://github.com/open-spaced-repetition/fsrs4anki/wiki/How-does-the-scheduler-work%3F#calculate-memory-states,

- Shouldn't the stability be equal to interval ÷ interval modifier?
https://github.com/open-spaced-repetition/fsrs-rs/blob/d389762b4fd73d9b43ab3af5f4f091807ddee96f/src/inference.rs#L88
- Shouldn't 0.1 * w10 be replaced by (1- R0) * w10? (R0 = request retention)
https://github.com/open-spaced-repetition/fsrs-rs/blob/d389762b4fd73d9b43ab3af5f4f091807ddee96f/src/inference.rs#L94

The above two points are based on a seemingly reasonable assumption that the request retention the user chooses would be similar to the retention achieved through SM-2. The current implementation is based on the assumption that SM-2 achieves 90% retention.

dae:
> reasonable assumption that the request retention the user chooses would be similar to the retention achieved through SM-2

That presumes the user previously was making use of the interval modifier, and I'm not sure that's a safe assumption: it was harder to manipulate, and now that FSRS provides a one-click button to suggest quite different numbers from 0.9, I fear the difference between actual and configured will tend to be larger than actual and 0.9.

We could add an extra preset variable that users could use to input their previous SM-2 retention, but given this only affects people with missing learning steps, I'm not sure it's worth the extra complexity. Wouldn't 0.9 be a close enough approximation?

dae:
Also, I based my code off the JS - I don't think it factors the interval multiplier into the conversion:

https://github.com/open-spaced-repetition/fsrs4anki/blob/dfaf660e5aa163402272a8d8b148869a45f09e52/fsrs4anki_scheduler.js#L232

user1823:
> That presumes the user previously was making use of the interval modifier,

I am sorry for the word choice, but I re-used the term that Jarrett used in the Wiki page. Here, interval modifier is not the same as the interval modifier in the deck config. Rather, it is just the ratio between the interval (given by FSRS) and the stability. Let me rename it as `retention factor`. It can be calculated as follows:
$$\text{Retention factor} = 9 \times \left(\frac{1}{\text{RequestRetention}} - 1 \right)$$

> now that FSRS provides a one-click button to suggest quite different numbers from 0.9, I fear the difference between actual and configured will tend to be larger than actual and 0.9.

This seems to be a reasonable concern. 

> We could add an extra preset variable that users could use to input their previous SM-2 retention, 

And this seems to be a reasonable solution. Probably, we can add it to an advanced section so that most users won't need to configure it.

>  I'm not sure it's worth the extra complexity. Wouldn't 0.9 be a close enough approximation?

Though this can't be generalized to all users, I still wanted to mention that my retention with SM-2 was about 94% and my requestRetention currently is also 0.94. And for this retention, the `retention factor` is about 0.57. So, the difference between 0.90 and 0.94 is huge.

L-M-Sherlock:
But we don't know the real retention for the interval scheduled by SM-2.

user1823:
> But we don't know the real retention for the interval scheduled by SM-2.

True, we don't know the retention for any specific card. But, we do know the average retention (from the stats). The user can enter the value by taking help from the stats.

user1823:
It is devs' choice.

dae:
As a partial revlog history [will now be used to determine memory state](https://github.com/open-spaced-repetition/fsrs-rs/issues/87#issuecomment-1741647208), the retention factor would only come into play when a card has a completely empty history. This feels like a rather niche case, and I'm not sure it's worth the increased UI complexity to solve. Anki won't write a memory state in such cases, so an add-on could easily fill in alternate values for such cases if you wished.

PR #89, opened 2023-09-28
wasm support

AlexErrant:
Do you have any interest in supporting wasm? This PR gets FSRS's `next_interval` working... I think. I haven't called it from Javascript yet - I just ran `wasm-pack build --target web && cat pkg/fsrs.d.ts;`, saw

```ts
export class FSRSwasm {
  free(): void;
/**
* @param {number | undefined} stability
* @param {number} desired_retention
* @param {number} rating
* @returns {number}
*/
  next_interval(stability: number | undefined, desired_retention: number, rating: number): number;
}
```

and declared victory.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -9,6 +9,7 @@ use burn::{
     module::{Module, Param},
     tensor::{backend::Backend, Data, Float, Shape, Tensor},
 };
+use wasm_bindgen::prelude::*;
 
 #[derive(Module, Debug)]
 pub struct Model<B: Backend> {
@@ -199,6 +200,21 @@ impl ModelConfig {
 pub struct FSRS<B: Backend = NdArrayBackend> {
     model: Option<Model<B>>,
     device: B::Device,
+}
+
+#[wasm_bindgen]
+pub struct FSRSwasm(FSRS<NdArrayBackend>);
+
+#[wasm_bindgen]
+impl FSRSwasm {
+    pub fn next_interval(
+        &self,
+        stability: Option<f32>,
+        desired_retention: f32,
+        rating: u32,
+    ) -> u32 {
+        self.0.next_interval(stability, desired_retention, rating)
+    }
 }
 
 impl FSRS<NdArrayBackend> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,7 +103,7 @@ checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
 [[package]]
 name = "burn"
 version = "0.10.0"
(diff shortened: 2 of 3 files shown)


src/model.rs
```diff
@@ -201,6 +202,21 @@ pub struct FSRS<B: Backend = NdArrayBackend> {
     device: B::Device,
 }
 
+#[wasm_bindgen]
+pub struct FSRSwasm(FSRS<NdArrayBackend>);
```
AlexErrant, line 206: `wasm_bindgen` [can't handle generics](https://github.com/rustwasm/wasm-bindgen/issues/3309), so I made it concrete.

src/model.rs
```diff
@@ -201,6 +202,21 @@ pub struct FSRS<B: Backend = NdArrayBackend> {
...
+impl FSRSwasm {
+    pub fn next_interval(
+        &self,
+        stability: Option<f32>,
+        desired_retention: f32,
+        rating: u32,
+    ) -> u32 {
+        self.0.next_interval(stability, desired_retention, rating)
+    }
+}
```
AlexErrant, line 218: Using the unfortunately verbose technique described [here](https://users.rust-lang.org/t/how-to-generate-wasm-for-generic-types-for-a-given-specific-type/67164). I am _very_ new to Rust, so there could be something obvious I'm missing. In other functional languages I'd use currying... but not sure if that's possible/reasonable here.

dae:
I'm afraid I don't see the point for such a trivial function that can easily be implemented in JS (and [already is](https://github.com/open-spaced-repetition/fsrs.js/blob/f8d3239d95d20fbea0141e27456a9e0d91047fda/src/index.ts#L312)). You also can't just change the library type, as that would prevent this library from being used in Rust programs - you should make a separate crate instead, which imports this one and exposes what you need.

AlexErrant:
I would like to expose other functions related to training as well, but kept the scope small.

> you should make a separate crate instead

Makes sense! Does this org have any interest in hosting the repo? Seems like The Place for FSRS stuff (but I won't be offended if you want _less_ responsibility).

L-M-Sherlock:
> Makes sense! Does this org have any interest in hosting the repo? Seems like The Place for FSRS stuff (but I won't be offended if you want less responsibility).

I'm glad to host it if you have implemented a useable version. I'm not familiar with wasm, so I don't know whether it's feasible.

By the way, will AnkiWeb support FSRS? @dae

dae:
It should already work, though I haven't integrated the tweaks made in the last few days yet.

AlexErrant:
> It should already work

FYI, maybe an issue with that? https://www.reddit.com/r/Anki/comments/16ucwhu/ankiweb_does_not_use_fsrs4/

Don't feel obligated to respond; I haven't experienced it myself.

dae:
They're using the custom scheduler version, not the built-in version.

Issue #88, opened 2023-09-28
Achieve parity with the Python optimizer


dae:
Currently the generated weights from this crate are slightly behind the Python optimizer. Not urgent, but in the long run, it would be nice if this crate could perform as well.

https://github.com/open-spaced-repetition/fsrs-benchmark#weighted-by-number-of-reviews

Any thoughts on what might be driving the differences?

L-M-Sherlock:
1. The current outlier filter only remove those outliers in pretrain-set. Py version remove the them in the train-set, too: https://github.com/open-spaced-repetition/fsrs-optimizer/blob/94be1a22d93e84e121186ea7181f55bc826639d3/src/fsrs_optimizer/fsrs_optimizer.py#L654-L664
2. The Py version split the trainset and optimizer the weights for each split and calculate the average weights as the final result: thttps://github.com/open-spaced-repetition/fsrs-optimizer/blob/94be1a22d93e84e121186ea7181f55bc826639d3/src/fsrs_optimizer/fsrs_optimizer.py#L1050-L1065
3. The Py version applied [additive smoothing](https://github.com/open-spaced-repetition/fsrs4anki/pull/358) to the pretrain: https://github.com/open-spaced-repetition/fsrs-optimizer/blob/94be1a22d93e84e121186ea7181f55bc826639d3/src/fsrs_optimizer/fsrs_optimizer.py#L847-L849

user1823:
> The current outlier filter only remove those outliers in pretrain-set. The py version removes the them in the train-set, too

In my opinion, the current outlier filter is too aggressive (it removes too many reviews). This is fine for the pretrain function (because the pretrain method is fragile). However, the train function should have access to more reviews. So, I think that we should build a less aggressive outlier filter for the train function.

There is no hurry in doing this. But, I wanted to make this point here so that you don't waste time and effort in applying the current outlier filter to the trainset.

user1823:
The python optimizer selects the parameters from the epoch with the minimum loss. I don't think that the rust optimizer behaves similarly.

Related commit: https://github.com/open-spaced-repetition/fsrs4anki/commit/1719daee92bd33bfbc65fe986ae0483e6299b074

L-M-Sherlock:
> The python optimizer selects the parameters from the epoch with the minimum loss. I don't think that the rust optimizer behaves similarly.

I also find out the difference, but the framework also doesn't also me to implement a similar feature.

dae:
[...]

PR #91, opened 2023-09-30
Run retention calculation in parallel

dae:
Roughtly 10x speed-up if you have enough cores on your CPU.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -774,6 +774,7 @@ dependencies = [
  "ndarray",
  "ndarray-rand",
  "rand",
+ "rayon",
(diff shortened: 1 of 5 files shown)


L-M-Sherlock (approved):
Pretty fast! It only costs me 1.23s to pass the test. Good job!

dae:
Currently weight training only uses one core. If point 2 on https://github.com/open-spaced-repetition/fsrs-rs/issues/88#issuecomment-1738493071 is implemented, we could potentially achieve similar performance improvements by using a separate core to calculate each of the values that we then average.

Issue #87, opened 2023-09-27
Use the first revlog in the "known" review history for converting SM-2 ivl & ease to memory states


user1823:
In Anki 23.10 beta 3, for cards with incomplete review history, Anki is using the _last_ revlog entry to calculate memory states from SM-2 ivl and ease. However, I recommend using the _first_ revlog entry. The reason is simple: The earlier we can calculate the memory states, the more accurate the results will be.

I described the algorithm for approaching such cards here: https://github.com/open-spaced-repetition/fsrs-rs/issues/63#issuecomment-1735323658

I implemented the same thing in the Helper add-on (with help from @L-M-Sherlock) in this PR: https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/248

Screenshots:

Memory states calculated using Anki:
<img width="540" alt="image" src="https://github.com/open-spaced-repetition/fsrs-rs/assets/92206575/78038734-efaa-41f1-a4dc-175ceffe0a4c">

Memory states calculated using Helper add-on
<img width="539" alt="image" src="https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/92206575/67e2f9a6-d3b7-4607-9c53-35bfd1e762fc">

user1823:
How does it make the converter too complex? Instead of using the last revlog entry, we are just using the first one and then the subsequent reviews of the same card are handled in the same way the reviews of normal cards are handled.

L-M-Sherlock:
Because it requires the full review history. The current method only requires the latest review log.

user1823:
> Because it requires the full review history. 

Calculation of memory states for the first time requires the full review history for all cards, not just these cards.

After each subsequent review, the memory states can be updated by using just the current review and the memory states after the last review, just like any other card.

Frankly, I don't understand how my proposed solution is different from the approach used for any other "normal" card except for using `memory_state_from_sm2` function instead of `init_states` function at the first revlog entry.

L-M-Sherlock:
Because the Rust code doesn't work in the same way as the helper add-on. The calculation of memory states and optimization share the same function. It's tricky to re-implement the same thing I implemented in the helper add-on.

user1823:
Ok, then let @dae have a look on this.

dae:
The revlog is not used when calculating memory state - only the card's current ease and interval are. The last revlog entry is used when answering cards, to determine the time elapsed since the last review.

[...]

PR #93, opened 2023-10-01
Support overriding initial S/D when revlog incomplete

dae:
https://github.com/open-spaced-repetition/fsrs-rs/issues/87

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -66,8 +66,8 @@ fn next_interval(stability: f32, request_retention: f32) -> u32 {
 impl<B: Backend> FSRS<B> {
     /// Calculate the current memory state for a given card's history of reviews.
     /// In the case of truncated reviews, [starting_state] can be set to the value of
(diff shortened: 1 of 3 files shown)


user1823:
@dae, thanks for implementing this. I just wanted to confirm one thing. After this change, the incomplete review history would be used for calculating the memory states but not for training. Right?

dae:
Correct.

L-M-Sherlock:
I have a point need to confirm: when the incomplete revlog has N revlog entry (e.g. 1...N), the first revlog will be converted to the start memory state, and the remaining revlogs' intervals and ratings are converted to t_history and r_history, right? The rating of the first review is irrelevant here.

dae:
I think this is something I'll need to handle on Anki's end - does the clarified docstring answer your question?

L-M-Sherlock (approved):
LGTM

PR #94, opened 2023-10-01
Laplace smoothing

asukaminato0721:
ref https://github.com/open-spaced-repetition/fsrs4anki/pull/358/files

also https://github.com/open-spaced-repetition/fsrs-rs/issues/88#issuecomment-1738493071


L-M-Sherlock (approved):
LGTM

PR #95, opened 2023-10-01
Feat/stratified k fold

L-M-Sherlock:
Port https://github.com/open-spaced-repetition/fsrs-optimizer/blob/fa63f61630a3e6a347e0576fc46b60e4561ac15e/src/fsrs_optimizer/fsrs_optimizer.py#L1049-L1082

also https://github.com/open-spaced-repetition/fsrs-rs/issues/88#issuecomment-1738493071

Consider to speed it up via https://github.com/open-spaced-repetition/fsrs-rs/pull/91#issuecomment-1741692297

By the way, the progress bar doesn't work as before. @dae

### DIFF
--- a/src/training.rs
+++ b/src/training.rs
@@ -207,8 +207,9 @@ impl<B: Backend> FSRS<B> {
         items: Vec<FSRSItem>,
         progress: Option<Arc<Mutex<ProgressState>>>,
     ) -> Result<Vec<f32>> {
+        let n_splits = 5;
         let average_recall = calculate_average_recall(&items);
-        let (pre_trainset, trainset) = split_data(items);
+        let (pre_trainset, trainsets) = split_data(items, n_splits);
         let initial_stability = pretrain(pre_trainset, average_recall)?;
         let config = TrainingConfig::new(
             ModelConfig {
@@ -218,14 +219,35 @@ impl<B: Backend> FSRS<B> {
             AdamConfig::new(),
         );
 
-        let model = train::<ADBackendDecorator<B>>(
-            trainset,
-            &config,
-            self.device(),
-            progress.map(ProgressCollector::new),
-        );
+        let mut weights_sets: Vec<Vec<f32>> = Vec::new();
 
-        Ok(model?.w.val().to_data().convert().value)
+        for i in 0..n_splits {
+            let trainset = trainsets
+                .iter()
+                .enumerate()
+                .filter(|&(j, _)| j != i)
+                .flat_map(|(_, trainset)| trainset.clone())
+                .collect();
+
+            let model = train::<ADBackendDecorator<B>>(
+                trainset,
+                &config,
+                self.device(),
+                progress.clone().map(ProgressCollector::new),
+            );
+            weights_sets.push(model?.w.val().to_data().convert().value)
+        }
+
+        let average_weights = weights_sets
+            .iter()
+            .fold(vec![0.0; weights_sets[0].len()], |sum, weights| {
+                sum.iter().zip(weights.iter()).map(|(a, b)| a + b).collect()
+            })
+            .iter()
+            .map(|&sum| sum / 5.0)
+            .collect();
+
+        Ok(average_weights)
     }
 }
 
@@ -335,9 +357,10 @@ mod tests {
             println!("Skipping test in CI");
             return;
         }
+        let n_splits = 5;
         let device = NdArrayDevice::Cpu;
         let items = anki21_sample_file_converted_to_fsrs();
-        let (pre_trainset, trainset) = split_data(items);
+        let (pre_trainset, trainsets) = split_data(items, n_splits);
         let average_recall = calculate_average_recall(&pre_trainset);
         let initial_stability = pretrain(pre_trainset, average_recall).unwrap();
         let config = TrainingConfig::new(
@@ -348,7 +371,29 @@ mod tests {
             AdamConfig::new(),
         );
 
-        let _model_trained =
-            train::<NdArrayAutodiffBackend>(trainset, &config, device, None).unwrap();
+        let mut weights_sets: Vec<Vec<f32>> = Vec::new();
+
+        for i in 0..n_splits {
+            let trainset = trainsets
+                .iter()
+                .enumerate()
+                .filter(|&(j, _)| j != i)
+                .flat_map(|(_, trainset)| trainset.clone())
+                .collect();
+            let model = train::<NdArrayAutodiffBackend>(trainset, &config, device, None);
+            weights_sets.push(model.unwrap().w.val().to_data().convert().value)
+        }
+
+        dbg!(&weights_sets);
+
+        let average_weights: Vec<f32> = weights_sets
+            .iter()
+            .fold(vec![0.0; weights_sets[0].len()], |sum, weights| {
+                sum.iter().zip(weights.iter()).map(|(a, b)| a + b).collect()
+            })
+            .iter()
+            .map(|&sum| sum / n_splits as f32)
+            .collect();
+        dbg!(average_weights);
     }
 }
--- a/src/training.rs
+++ b/src/training.rs
@@ -174,7 +174,7 @@ impl DashboardRenderer for ProgressCollector {
 pub(crate) struct TrainingConfig {
     pub model: ModelConfig,
     pub optimizer: AdamConfig,
-    #[config(default = 16)]
+    #[config(default = 216)]
     pub num_epochs: usize,
     #[config(default = 1024)]
     pub batch_size: usize,
@@ -347,6 +347,7 @@ mod tests {
     use crate::pre_training::pretrain;
     use burn::backend::ndarray::NdArrayDevice;
     use burn::backend::NdArrayAutodiffBackend;
+    use rayon::prelude::IntoParallelIterator;
 
     #[test]
     fn training() {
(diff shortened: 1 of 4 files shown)


src/training.rs
```diff
@@ -348,7 +368,29 @@ mod tests {
...
+        for i in 0..n_splits {
+            let trainset = trainsets
+                .par_iter()
+                .enumerate()
+                .filter(|&(j, _)| j != i)
+                .flat_map(|(_, trainset)| trainset.clone())
+                .collect();
+            let model = train::<NdArrayAutodiffBackend>(trainset, &config, device, None);
+            weights_sets.push(model.unwrap().w.val().to_data().convert().value)
+        }
```
dae, line 382: This is roughly twice as fast:

```rust
        let weights_sets: Vec<Vec<f32>> = (0..n_splits).into_par_iter().map(|i| {
            let trainset = trainsets
                .iter()
                .enumerate()
                .filter(|&(j, _)| j != i)
                .flat_map(|(_, trainset)| trainset.clone())
                .collect();
            let model = train::<NdArrayAutodiffBackend>(trainset, &config, device, None);
            model.unwrap().w.val().to_data().convert().value
        }).collect();
```

It's still about 2x slower than the original implementation before this PR though, because there's now 4x as much work to be done, as the trainsets are being added together. Would it make sense to only use one of the trainsets instead?

```rust
        let weights_sets: Vec<Vec<f32>> = (0..n_splits).into_par_iter().map(|i| {
            let trainset = trainsets[i].clone();
            let model = train::<NdArrayAutodiffBackend>(trainset, &config, device, None);
            model.unwrap().w.val().to_data().convert().value
        }).collect();
```

dae:
> the progress bar doesn't work as before

We won't be able to use the built-in progress bar - I'll need to rewrite the progress code to handle the fact that training is happening in multiple threads at once.

dae:
There is an issue in burn that is preventing us from getting optimum performance. I've sent them a PR: https://github.com/burn-rs/burn/pull/839

Currently n_splits is hard-coded to 5, which means that on CPUs with more cores, we wouldn't be fully taking advantage of them. Would it make sense to allow splits to go up to 16? If so, I can do that in a follow-up PR that determines the splits based on available CPU cores.

src/training.rs
```diff
@@ -174,7 +174,7 @@ impl DashboardRenderer for ProgressCollector {
 pub(crate) struct TrainingConfig {
     pub model: ModelConfig,
     pub optimizer: AdamConfig,
-    #[config(default = 16)]
+    #[config(default = 216)]
```
dae, line 177: Was this deliberate? With such a big jump in epochs, we'll lose any performance gains we've made here.

dae:
split=16
metric: LogLoss
FSRS-rs mean: 0.3899
metric: RMSE
FSRS-rs mean: 0.3345
metric: RMSE(bins)
FSRS-rs mean: 0.0624

split=5
metric: LogLoss
FSRS-rs mean: 0.3894
metric: RMSE
FSRS-rs mean: 0.3343
metric: RMSE(bins)
FSRS-rs mean: 0.0612

split=1
metric: LogLoss
FSRS-rs mean: 0.3878
metric: RMSE
FSRS-rs mean: 0.3337
metric: RMSE(bins)
FSRS-rs mean: 0.0598

Maybe we could look at the size of the dataset, and use that to help decide how many splits? Eg try to avoid creating splits with less than 5,000 items.

L-M-Sherlock:
> Currently n_splits is hard-coded to 5

Because the py optimizer applies `n_splits=5`. I want to compare them in the benchmark with the same hype-parameters.

dae:
I tried with: `let n_splits = 16.min(items.len() / 5000).max(1);`, and it looks like we get the best of both worlds: better performance, and weights equal to the split=1 case:

metric: LogLoss
FSRS-rs mean: 0.3877
metric: RMSE
FSRS-rs mean: 0.3337
metric: RMSE(bins)
FSRS-rs mean: 0.0597

What if we adjusted the Python optimizer to use the same logic in calculating splits, so you could still compare the two? In Anki I think 16 should be detected from the current cpu's core count, but we could an env var to force it to be a specific number, so the tests are consistent.

dae:
Also there's currently one inconsistency: revlog_history-60d332b.json is generated for FSRS-rs but not the others for some reason. When removed, things improve slightly more:

metric: LogLoss
FSRS-rs mean: 0.3859
metric: RMSE
FSRS-rs mean: 0.3328
metric: RMSE(bins)
FSRS-rs mean: 0.0587

L-M-Sherlock:
Here is the result of the latest commit:

Model: FSRS-rs
Total number of users: 71
Total number of reviews: 4632965
metric: LogLoss
FSRS-rs mean: 0.3858
metric: RMSE
FSRS-rs mean: 0.3327
metric: RMSE(bins)
FSRS-rs mean: 0.0594

I will test `let n_splits = 16.min(items.len() / 5000).max(1);`.

dae:
It seems 1000 reviews is not enough:

Model: FSRS-rs
Total number of users: 71
Total number of reviews: 4632965
metric: LogLoss
FSRS-rs mean: 0.3874
metric: RMSE
FSRS-rs mean: 0.3334
metric: RMSE(bins)
FSRS-rs mean: 0.0613

Will try with 10,000 for comparison.

dae:
By the way, AnkiWeb's terms and conditions allow us to extract revlog data from AnkiWeb users. I could potentially build a dataset much bigger than 71 if that were helpful (though it may take me a while to do).

dae:
10,000 review split:

Total number of users: 71
Total number of reviews: 4632965
metric: LogLoss
FSRS-rs mean: 0.3858
metric: RMSE
FSRS-rs mean: 0.3327
metric: RMSE(bins)
FSRS-rs mean: 0.0586

L-M-Sherlock:
Weird. Usually the benchmark only costs me 15 mins, but it costs almost one hours in this time.

dae:
The change I suggested also needs to be applied to the non-test case:

```patch
diff --git a/src/training.rs b/src/training.rs
index 14e4d03..a8c59c7 100644
--- a/src/training.rs
+++ b/src/training.rs
@@ -18,7 +18,7 @@ use burn::train::{ClassificationOutput, TrainOutput, TrainStep, TrainingInterrup
 use burn::{config::Config, module::Param, tensor::backend::ADBackend, train::LearnerBuilder};
 use core::marker::PhantomData;
 use log::info;
-use rayon::prelude::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
+use rayon::prelude::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator, IntoParallelIterator};
 use std::path::Path;
 use std::sync::{Arc, Mutex};

@@ -206,7 +206,8 @@ impl<B: Backend> FSRS<B> {
         items: Vec<FSRSItem>,
         progress: Option<Arc<Mutex<ProgressState>>>,
     ) -> Result<Vec<f32>> {
-        let n_splits = 5;
+        let n_splits = 16.min(items.len() / 10000).max(1);
+        dbg!(n_splits);
         let average_recall = calculate_average_recall(&items);
         let (pre_trainset, trainsets) = split_data(items, n_splits);
         let initial_stability = pretrain(pre_trainset, average_recall)?;
@@ -218,24 +219,17 @@ impl<B: Backend> FSRS<B> {
             AdamConfig::new(),
         );

-        let mut weights_sets: Vec<Vec<f32>> = Vec::new();
-
-        for i in 0..n_splits {
-            let trainset = trainsets
-                .par_iter()
-                .enumerate()
-                .filter(|&(j, _)| j != i)
-                .flat_map(|(_, trainset)| trainset.clone())
-                .collect();

+        let weights_sets: Vec<Vec<f32>> = (0..n_splits).into_par_iter().map(|i| {
+            let trainset = trainsets[i].clone();
             let model = train::<ADBackendDecorator<B>>(
                 trainset,
                 &config,
                 self.device(),
                 progress.clone().map(ProgressCollector::new),
             );
-            weights_sets.push(model?.w.val().to_data().convert().value)
-        }
+            model.unwrap().w.val().to_data().convert().value
+        }).collect();

         let average_weights = weights_sets
             .iter()
```

dae:
10000 should be 5000

dae:
Oh, and you also need to be using the changes in https://github.com/burn-rs/burn/pull/839 to get optimum performance. It might be worth waiting a day or two to see if that gets merged.

L-M-Sherlock:
> ```diff
> +            let trainset = trainsets[i].clone();
> ```

It's inconstant with the previous code.

dae:
https://github.com/open-spaced-repetition/fsrs-rs/pull/95#discussion_r1342201925

L-M-Sherlock:
[KFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold) divides all the samples in $k$
 groups of samples, called folds (if $k=n$, this is equivalent to the Leave One Out strategy), of equal sizes (if possible). The prediction function is learned using $k-1$ folds, and the fold left out is used for test.

L-M-Sherlock:
The verification items will be used for selecting the best weights when we implement https://github.com/open-spaced-repetition/fsrs-rs/issues/88#issuecomment-1742424027

L-M-Sherlock:
n_splits = 4:

Model: FSRS-rs
Total number of users: 71
Total number of reviews: 4632965
metric: LogLoss
FSRS-rs mean: 0.3859
metric: RMSE
FSRS-rs mean: 0.3327
metric: RMSE(bins)
FSRS-rs mean: 0.0593

L-M-Sherlock:
pub(crate) struct TrainingConfig {
    pub model: ModelConfig,
    pub optimizer: AdamConfig,
    #[config(default = 5)]
    pub num_epochs: usize,
    #[config(default = 512)]
    pub batch_size: usize,
    #[config(default = 1)]
    pub num_workers: usize,
    #[config(default = 42)]
    pub seed: u64,
    #[config(default = 4e-2)]
    pub learning_rate: f64,
}

n_splits = 5

Model: FSRS-rs
Total number of users: 71
Total number of reviews: 4632965
metric: LogLoss
FSRS-rs mean: 0.3853
metric: RMSE
FSRS-rs mean: 0.3325
metric: RMSE(bins)
FSRS-rs mean: 0.0582

It seems to be the best result.

asukaminato0721 (approved):

PR #96, opened 2023-10-01
Fix/BatchShuffledDataLoader

L-M-Sherlock:
I find out a problem with burn:
- https://github.com/burn-rs/burn/pull/703

It calls `ShuffledDataset` to reshuffle the dataset at each epoch. But it breaks the BatchShuffledDataset. So I implemented a BatchShuffledDataLoader to solve this problem. Maybe letting the burn support BatchShuffledDataset is a better solution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,8 @@ rand = "0.8.5"
 rayon = "1.8.0"
 serde = "1.0.188"
 snafu = "0.7.5"
-strum = { version = "0.25.0", features = ["derive"] }
 spin = { version = "0.9.8", features = ["mutex", "spin_mutex"] }
+strum = { version = "0.25.0", features = ["derive"] }
 
 [dev-dependencies]
 chrono = { version = "0.4.26", default-features = false, features = ["std", "clock"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -30,6 +30,7 @@ rand = "0.8.5"
 rayon = "1.8.0"
 serde = "1.0.188"
 snafu = "0.7.5"
+spin = { version = "0.9.8", features = ["mutex", "spin_mutex"] }
```
dae, line 33: Burn uses this because they also need to operate on embedded devices, but that's not a concern for us. I'd just use the standard Mutex available in Rust, instead of relying on an external crate.

src/batch_shuffle.rs
```diff
@@ -33,7 +45,7 @@ where
             let end_index = (start_index + batch_size).min(len);
             indices.extend(start_index..end_index);
         }
-
+        info!("indices: {:?}", &indices);
```
dae, line 48: Please remove these info! lines before commiting.

dae (commented):
Mostly looks good, just a few comments:

dae (approved):

PR #103, opened 2023-10-13
Fix/Use power forgetting curve in memory_state_from_sm2

user1823:
FSRS v4 uses a power forgetting curve. The earlier formula was based on the exponential forgetting curve.

I made the same mistake in the helper add-on code, and I think that dae picked up the formula from there.

The correct formula written as Math can be seen here: https://github.com/open-spaced-repetition/fsrs4anki/wiki/How-does-the-scheduler-work%3F#interval-to-stability

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -95,7 +95,7 @@ impl<B: Backend> FSRS<B> {
         interval: f32,
         sm2_retention: f32,
     ) -> MemoryState {
-        let stability = interval.max(0.1) / (9 * (1 / sm2_retention - 1));
+        let stability = interval.max(0.1) / (9.0 * (1.0 / sm2_retention - 1.0));
         let w = &self.model().w;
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
I think it is incorrect to use power forgetting curve here. Because the forgetting curve of SM-2 is  $R = e^{\ln(0.9) \cdot \cfrac{I}{S}}$. We shouldn't use the forgetting curve of FSRS in SM-2 convertor.

user1823:
Strictly speaking, SM-2 doesn't have the concept of stability and retrievability. So, it doesn't seem correct to say that SM-2 uses the exponential forgetting curve.

Nonetheless, we are trying to get an estimate of the "FSRS stability" that will produce a given retention at a given interval. So, in my opinion, we should use the FSRS forgetting curve here.

L-M-Sherlock (approved):

PR #108, opened 2023-10-22
Update burn-rs

dae:
Should enable us to implement early stopping:

https://github.com/burn-rs/burn/blob/af813d09ed2ee33ffcf13dd1e2ec127fb8d722dd/examples/mnist/src/training.rs#L74


L-M-Sherlock (approved):
LGTM

PR #111, opened 2023-10-26
Feat/calc SInc based on last d

L-M-Sherlock:


dae:
If I've understood correctly, this looks like a rather minor improvement - is it ok to skip this for the 23.10 release and include it in the next patch update instead? I'm nervous about pulling in changes this close to a release.

L-M-Sherlock:
Feel free to delay it. I haven't modified the add-on and run the benchmark. Only one thing should be noticed: the model's structure is changed, so the user should re-optimize the parameters when the next patch releases.

asukaminato0721 (approved):

asukaminato0721 (approved):

PR #122, opened 2023-11-22
Feat/speed up finding optimal retention via brent's method

L-M-Sherlock:
Brent's method is faster and more accurate:
1. https://github.com/open-spaced-repetition/fsrs-optimizer/issues/60#issuecomment-1812185306
2. https://github.com/open-spaced-repetition/fsrs-optimizer/issues/60#issuecomment-1816575028

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,6 +5,7 @@ pub enum FSRSError {
     NotEnoughData,
     Interrupted,
     InvalidWeights,
+    OptimalNotFound,
 }
 
 pub type Result<T, E = FSRSError> = std::result::Result<T, E>;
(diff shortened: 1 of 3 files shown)


dae:
My review is focused on the Rust code, and not the math itself. Looks fine to me, but presumably we'll need to update the progress tracking code to capture the time spent in bracket(). Out of curiosity, how does a before/after benchmark perform?

L-M-Sherlock:
> Out of curiosity, how does a before/after benchmark perform?

In debug mode, the test costs 38s for the old method. The new method costs 17s.

dae:
Nice!

asukaminato0721 (approved):

PR #129, opened 2023-12-11
Increase the sample size.rs

Expertium:
Since the new method is roughly 2 times faster, how about sacrificing that speed in favor of accuracy? Of course, ideally we want both, but if I had to choose, I would say accuracy is more important in this case.
After this change, the speed should be more or less the same as with the old method.

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -395,15 +395,15 @@ where
     const GOLD: f64 = 1.618_033_988_749_895; // wait for https://doc.rust-lang.org/std/f64/consts/constant.PHI.html
     const MAXITER: i32 = 20;
 
-    let mut fa = -sample(config, weights, xa, 5, progress)?;
-    let mut fb = -sample(config, weights, xb, 5, progress)?;
+    let mut fa = -sample(config, weights, xa, 10, progress)?;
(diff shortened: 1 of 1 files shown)


L-M-Sherlock (approved):

Expertium:
@L-M-Sherlock I think there is an error in this line:
`let (mut a, mut b) = (xa.min(xc), xa.max(xc));`
Original Python code:
```
    if xa < xc:
        a = xa
        b = xc
    else:
        a = xc
        b = xa
```

izik1:
that looks correct and equivalent to:
```rs
let (mut a, mut b) = if xa < xc { (xa, xc) } else { (xc, xa) };
```

L-M-Sherlock:
> `let (mut a, mut b) = (xa.min(xc), xa.max(xc));`

This code is correct. a is the minimum between xa and xc, and b is the maximum between xa and xc.

Expertium:
By the way, how can I tell whether this is implemented in the new beta?

dae:
Check which fsrs-rs commit Anki currently points to: https://github.com/ankitects/anki/blob/edd38ca06730d7fc16804f52ce10f6bc54c3d145/Cargo.toml#L39

PR #134, opened 2023-12-16
Feat/flat power forgetting curve

L-M-Sherlock:
Idea: https://github.com/open-spaced-repetition/fsrs4anki/issues/461#issuecomment-1787935005

Python Implementation: https://github.com/open-spaced-repetition/fsrs-optimizer/commit/8ac5d5bf763b2fb72857ffa3342e9b2b3a1d0d91

Help wanted: There are duplicated function named `power_forgetting_curve`. Is it possible to refactor them into one function?

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -410,7 +410,7 @@ mod tests {
         let metrics = fsrs.evaluate(items.clone(), |_| true).unwrap();
 
         Data::from([metrics.log_loss, metrics.rmse_bins])
-            .assert_approx_eq(&Data::from([0.20457467, 0.02268843]), 5);
+            .assert_approx_eq(&Data::from([0.21364396810531616, 0.05370686203241348]), 5);
 
         let fsrs = FSRS::new(Some(WEIGHTS))?;
         let metrics = fsrs.evaluate(items, |_| true).unwrap();
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -12,9 +12,9 @@ use crate::model::Model;
 use crate::training::BCELoss;
 use crate::{FSRSError, FSRSItem};
 use burn::tensor::ElementConversion;
-pub const DECAY: f64 = -0.5;
+pub(crate) const DECAY: f64 = -0.5;
 /// (9/10) ^ (1 / DECAY) - 1
-pub const FACTOR: f64 = 19f64 / 81f64;
+pub(crate) const FACTOR: f64 = 19f64 / 81f64;
 /// This is a slice for efficiency, but should always be 17 in length.
 pub type Weights = [f32];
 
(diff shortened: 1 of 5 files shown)


asukaminato0721:
linux:

```
---- pre_training::tests::test_loss stdout ----
thread 'pre_training::tests::test_loss' panicked at src/pre_training.rs:327:9:
assertion `left == right` failed
  left: 14.5771
 right: 14.577101
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

win:

```
---- pre_training::tests::test_loss stdout ----
thread 'pre_training::tests::test_loss' panicked at src\pre_training.rs:327:9:
assertion `left == right` failed
  left: 14.5771
 right: 14.577101
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

floating point again :(

asukaminato0721 (approved):

dae:
I can't see an easy way to combine the three power_forgetting_curves, as you have one that uses raw floats, one that uses NdArray, and one that uses tensors. You could potentially just have a raw float one and then convert the tensors to floats and back again, but it might be a bit slower if it's called frequently.

src/inference.rs
```diff
@@ -12,7 +12,9 @@ use crate::model::Model;
 use crate::training::BCELoss;
 use crate::{FSRSError, FSRSItem};
 use burn::tensor::ElementConversion;
-
+pub const DECAY: f64 = -0.5;
```
dae, line 15: pub(crate) would be a better choice

asukaminato0721:
> Is it possible to refactor them into one function?

They have different types, not something like `Tensor<B, 1>` vs `Tensor<B, 2>`

instead, `&Array1` vs `Array1` vs `f32` vs `f64` vs `Tensor`, so it's hard to use `trait` to merge them.

Maybe first we need to unify the array type...

asukaminato0721:
> > Is it possible to refactor them into one function?
> 
> They have different types, not something like `Tensor<B, 1>` vs `Tensor<B, 2>`
> 
> instead, `&Array1` vs `Array1` vs `f32` vs `f64` vs `Tensor`, so it's hard to use `trait` to merge them.
> 
> Maybe first we need to unify the array type...

Some small thought, maybe (I am not sure how to write the trait, this is a draft.)

```rs
fn power_forgetting_curve(a: T1, b: T2) -> T1
where
    T1: Add<Rhs = xxx, Ouput = T1> + Mul<Rhs = xxx, Output = T1> + Pow<Rhs = xxx, Output = T1>,
    T2: Add<Rhs = xxx, Ouput = T2> + Mul<Rhs = xxx, Output = T2> + Pow<Rhs = xxx, Output = T2>,
{
    a.some_operator(b)
}
```

PR #139, opened 2023-12-19
Fix/sort init_s0 on generating from r_s0_default

L-M-Sherlock:


asukaminato0721:
There is always a thought: can this `rating_stability` be `[Option<f32>; 5]` ?

asukaminato0721 (approved):

PR #154, opened 2024-02-01
Feat/compare current weights with another

L-M-Sherlock:
Context:
- https://github.com/ankitects/anki/issues/2972#issuecomment-1920719942

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -475,7 +475,7 @@ mod tests {
             .compare_with_weights(items, &DEFAULT_WEIGHTS, |_| true)
             .unwrap();
 
-        assert_eq!(self_is_better, true);
+        assert!(self_is_better);
         Data::from([self_by_other, other_by_self])
             .assert_approx_eq(&Data::from([0.015_987_674, 0.019_702_684]), 5);
         Ok(())
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -274,7 +274,7 @@ impl<B: Backend> FSRS<B> {
             let (_state, retention) = infer::<B>(model_other, batch.clone());
             let pred = retention.clone().to_data().convert::<f32>().value;
             all_predictions_other.extend(pred);
-            
+
             let true_val = batch.labels.clone().to_data().convert::<f32>().value;
             all_true_val.extend(true_val);
             progress_info.current += chunk.len();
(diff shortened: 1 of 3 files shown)


L-M-Sherlock:
> It would probably be cleaner to use the weights that are passed in at model creation though

Do you mean the function accept two weights instead of one weights?

dae:
I'm sorry Jarrett, I did not read your code carefully before commenting, so my comments were about what I assumed the code was doing, instead of what it was actually doing.

Having looked into it more carefully today, it looks like you intend it to be used like so:

- Caller creates a FSRS struct and calls .compute_weights(items)  to calculate the new weights
- Caller creates an FSRS struct with the old weights, and passes the returned weights into .compare_with_weights() to see if the new weights are better or not.

Is that correct?

src/inference.rs
```diff
@@ -236,6 +237,57 @@ impl<B: Backend> FSRS<B> {
     pub fn current_retrievability(&self, state: MemoryState, days_elapsed: u32) -> f32 {
         (days_elapsed as f32 / (state.stability * 9.0) + 1.0).powi(-1)
     }
+
+    /// Compare the current model and weights with another set of weights.
+    /// Returns a tuple with a boolean indicating whether the current model is better than the other,
```
dae, line 242: I think it would be clearer if the returned bool was true if the new weights are better, and then the docs could be updated:

```suggestion
    /// Returns a tuple with a boolean that's true if the provided weights are better
```

Alternatively, since you're already returning the two universal values, perhaps the boolean is not required, and the docstring could be updated instead, to something like:

> Returns the universal metrics for the existing and provided parameters. If the first value is smaller than the second value, the provided parameters are worse than the previous ones.

L-M-Sherlock: In my view, the suggested comment is confusing, because I don't know what the provided weights means. Does it mean the input parameters? If so, the comment is wrong because the boolean is true when the input parameters are worse than the parameters of the model itself.

L-M-Sherlock: > Returns the universal metrics for the existing and provided parameters. If the first value is smaller than the second value, the provided parameters are worse than the previous ones.

I prefer this one. I will update the code.

dae (approved):
Approving with optional extra suggestion: rename to compare_with_parameters(), and change `weights:` to `parameters:`. We can follow up with a future PR that does a s/weights/parameters for consistency.

dae:
On second thought, `universal_metrics` might be better than `compare_with_parameters`, since the caller is expected to do the comparison. Up to you :-)

PR #162, opened 2024-03-05
Fix/early stop

L-M-Sherlock:


asukaminato0721 (approved):

dae:
I'm curious about this change. Doesn't this now allow the regressions for config.num_epochs? How is that different from turning off the early stopping feature?

L-M-Sherlock:
I'm implementing a custom training loop to select the best parameters after the all epochs are complete. The early stop doesn't behave as I expected.

L-M-Sherlock:
https://github.com/open-spaced-repetition/fsrs-rs/pull/163

PR #164, opened 2024-03-06
add hook reminds

asukaminato0721:
Will reduce some CI format error.

because of <https://stackoverflow.com/questions/427207/can-git-hook-scripts-be-managed-along-with-the-repository/427269#427269>, can't add it into repo directly.


L-M-Sherlock (approved):
LGTM

PR #165, opened 2024-03-10
Feat/minimal workload for optimal retention

L-M-Sherlock:
- https://github.com/open-spaced-repetition/optimal-desired-retention/blob/main/notebook.ipynb
- https://github.com/open-spaced-repetition/fsrs-optimizer/pull/89

I refactor the approach to calculate the optimal retention. Now the optimizer will output the optimal retention to minimize the workload/cost per memorization.

image

Ideally, the user only needs to input `learn_span` in the new optimizer for retention. So it's easier to use than the previous implementation.

image

By the way, we may notice users that desired retention <  optimal retention is worse than desired retention > optimal retention. Because the former means **less** memorization & **more** workload and the latter means **more** memorization & **more** workload.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1055,7 +1055,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "burn",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1055,7 +1055,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.5.1"
+version = "0.5.2"
 dependencies = [
  "burn",
  "chrono",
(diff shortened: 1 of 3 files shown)


asukaminato0721 (approved):

dae:
So we can remove deck size and minutes/day, but I presume we still need to expose the daily new and review limits?

dae:
Sorry, to confirm, do you mean we should not add either? Or do you mean we should just add the new limit?

L-M-Sherlock:
We only need `Days to simulate`. I have made necessary modifications in Anki's part in the PR.

dae:
This is still in simulator config:

```
    pub learn_limit: usize,
    pub review_limit: usize,
```

Would it make sense to remove these in the future? (not urgent)

L-M-Sherlock:
> Would it make sense to remove these in the future? (not urgent)

Yeah it's still in simulator config because they are useful if we want to provide built-in simulator in the future. They are just useless in optimizing the retention.

PR #166, opened 2024-03-10
Fix/progress goes backwards

L-M-Sherlock:
fix https://forums.ankiweb.net/t/anki-24-04-beta-rc/41792/53?u=l.m.sherlock

In this PR, we will initialize the progress before training.


asukaminato0721 (approved):

dae:
Thanks for looking into this!

PR #169, opened 2024-03-16
export simulate & clippy nursery

L-M-Sherlock:
I plan to implement the simulator in Anki, so I need to export `simulate` at first.

PR #170, opened 2024-03-16
Fix/add pub(crate) back

L-M-Sherlock:
Discussed in https://github.com/open-spaced-repetition/fsrs-rs/pull/169#issuecomment-2001917206

### DIFF
--- a/src/batch_shuffle.rs
+++ b/src/batch_shuffle.rs
@@ -13,7 +13,7 @@ use std::{
 
 use crate::{dataset::FSRSDataset, FSRSItem};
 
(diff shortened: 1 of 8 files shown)


dae:
If check.sh needs updating, you can use

cargo clippy -- -Dclippy::nursery -Aclippy::redundant_pub_crate

L-M-Sherlock:
> If check.sh needs updating, you can use
> 
> cargo clippy -- -Dclippy::nursery -Aclippy::redundant_pub_crate

I haven't add `clippy::nursery` to check.sh, so we can merge it now.

dae (approved):

PR #172, opened 2024-03-30
Feat/add rules for returning parameters

L-M-Sherlock:
Context: https://github.com/ankitects/anki/issues/3094#issuecomment-2027764614

### DIFF
--- a/src/training.rs
+++ b/src/training.rs
@@ -501,26 +501,6 @@ mod tests {
         dbg!(average_recall);
         let initial_stability = pretrain(pre_trainset, average_recall).unwrap();
         dbg!(initial_stability);
-        let pretrained_parameters: Vec<f32> = initial_stability
-            .into_iter()
-            .chain(DEFAULT_PARAMETERS[4..].iter().copied())
-            .collect();
-        let pretrained_fsrs = FSRS::new(Some(&pretrained_parameters)).unwrap();
-        let pretrained_rmse = pretrained_fsrs
-            .evaluate(testset.clone(), |_| true)
-            .unwrap()
-            .rmse_bins;
-        let default_fsrs = FSRS::new(Some(&[])).unwrap();
-        let default_rmse = default_fsrs
-            .evaluate(testset.clone(), |_| true)
-            .unwrap()
-            .rmse_bins;
-        let (best_parameters, best_rmse) = if pretrained_rmse < default_rmse {
-            (pretrained_parameters.clone(), pretrained_rmse)
-        } else {
-            (DEFAULT_PARAMETERS.to_vec(), default_rmse)
-        };
-        dbg!(best_parameters);
         let config = TrainingConfig::new(
             ModelConfig {
                 freeze_stability: true,
@@ -591,7 +571,6 @@ mod tests {
             .evaluate(testset, |_| true)
             .unwrap()
             .rmse_bins;
-        dbg!(best_rmse);
         dbg!(optimized_rmse);
     }
 }
--- a/src/training.rs
+++ b/src/training.rs
@@ -222,13 +222,33 @@ impl<B: Backend> FSRS<B> {
             finish_progress();
             e
         })?;
-        let pretrained_parameters = initial_stability
+        let pretrained_parameters: Vec<f32> = initial_stability
             .into_iter()
             .chain(DEFAULT_PARAMETERS[4..].iter().copied())
             .collect();
-        if pre_trainset.len() + testset.len() < 32 || testset.is_empty() {
+        if testset.is_empty() {
             finish_progress();
             return Ok(pretrained_parameters);
+        }
+
+        let pretrained_fsrs = FSRS::new(Some(&pretrained_parameters)).unwrap();
+        let pretrained_rmse = pretrained_fsrs
+            .evaluate(testset.clone(), |_| true)
+            .unwrap()
+            .rmse_bins;
+        let default_fsrs = FSRS::new(None).unwrap();
+        let default_rmse = default_fsrs
+            .evaluate(testset.clone(), |_| true)
+            .unwrap()
+            .rmse_bins;
+        let (best_parameters, best_rmse) = if pretrained_rmse < default_rmse {
+            (pretrained_parameters.clone(), pretrained_rmse)
+        } else {
+            (DEFAULT_PARAMETERS.to_vec(), default_rmse)
+        };
+        if pre_trainset.len() + testset.len() < 32 {
+            finish_progress();
+            return Ok(best_parameters);
         }
 
         let config = TrainingConfig::new(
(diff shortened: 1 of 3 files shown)


dae:
I was thinking of doing a stable release later today or tomorrow. I'm ok with including this change in the stable build, if you are reasonably confident it won't cause a regression / doesn't need any testing.

Expertium:
@L-M-Sherlock previously, you said "It's unnecessary to evaluate the default parameters. For all new users, the old_parameters is default_parameters". But that is true only for the very first optimization. It's not true for consecutive optimizations, so comparison with default parameters should be added.

L-M-Sherlock:
> But that is true only for the very first optimization. It's not true for consecutive optimizations

Do you mean the optimized parameters become worse than the default parameters over time?

By the way, each evaluation slows down the optimization.

Expertium:
> Do you mean the optimized parameters become worse than the default parameters over time?

Yes, it can happen sometimes. According to Rlz:

Size: 16384
Amt where 'default' does better than 'opt': 61 / 3299
Mean RMSE: 0.089 (default) vs 0.093 (opt)

Size: 32768
Amt where 'default' does better than 'opt': 34 / 2119
Mean RMSE: 0.107 (default) vs 0.112 (opt)

So even for very large sample sizes, default parameters are better 1-2% of the time.

user1823:
> I'm ok with including this change in the stable build, if you are reasonably confident it won't cause a regression / doesn't need any testing.

I think that doing some more testing would be better. In addition, it would be better to gather feedback from users about things like whether this increases the optimization time significantly with their collection and hardware.

So, I would suggest releasing 24.04 as it is. This change can be included in the next release.

> So even for very large sample sizes, default parameters are better 1-2% of the time.

In such cases, would it be better to use the optimized values of first 4 parameters and default values of the rest rather than using the default values of all parameters?

Expertium:
> In such cases, would it be better to use the optimized values of first 4 parameters and default values of the rest rather than using the default values of all parameters?

The rule compares all 3: default, pretrain and full. Of course, a situation where pretrain is better than full but worse than default would be exceptionally rare.

user1823:
> Of course, a situation where pretrain is better than full but worse than default would be exceptionally rare.

Did Rlz find such cases during the analysis? If there were very few cases, we can skip comparing against the default parameters. This would atleast decrease the time taken for optimization. In addition, this can prevent the confusion arising from a sudden shift to the default parameters.

> This could cause confusion if the user is paying close attention to the values of parameters.

Not only if the user is paying attention to the parameters.
The parameters (especially the initial stability) affect the memory states significantly. So, jumping from optimized parameters to default parameters may significantly affect the next intervals of the cards.

Expertium:
Size: 16384
Amt where 'default' does better than 'opt': 61 / 3299
Mean RMSE: 0.089 (default) vs 0.093 (opt)
Amt where 'opt' does better than 'default': 3238 / 3299
Mean RMSE: 0.078 (opt) vs 0.104 (default)
Amt where 'default' does better than 'pretrain': 338 / 3299
Mean RMSE: 0.094 (default) vs 0.096 (pretrain)
Amt where 'pretrain' does better than 'default': 2959 / 3299
Mean RMSE: 0.091 (pretrain) vs 0.105 (default)
Amt where 'opt' does better than 'pretrain': 3205 / 3299
Mean RMSE: 0.078 (opt) vs 0.092 (pretrain)
Amt where 'pretrain' does better than 'opt': 94 / 3299
Mean RMSE: 0.093 (pretrain) vs 0.094 (opt)

user1823:
This doesn't answer the specific question. I asked for the cases where the RMSE is in the order:
 
def < pretrain < opt

Answering the question requires looking at the raw data. This summary is not that useful.

Also, what about allowing small regressions in RMSE, considering that RMSE may not be the "perfect" metric?

Expertium:
@RlzHi could you please analyze all possible permutations of these 3 losses, to see how often certain permutations occur?

For example, how often default < pretrain < opt occurs, or how often pretrain < default < opt occurs

L-M-Sherlock:
I find a new problem: without the minimum limit, the user will only get this error:

image

Because the first review entry of new cards are removed, so the number of reviews for optimization is zero.

Expertium:
Just use the default parameters

L-M-Sherlock:
So, should we remove this error?

Expertium:
Yes

L-M-Sherlock:
But how should the `evaluate` feature behave if the number of reviews is zero?

Expertium:
Evaluate default parameters, since that's what will be used

L-M-Sherlock:
> Evaluate default parameters, since that's what will be used

But the size of testset is zero. We can't evaluate the parameters in an empty testset.

user1823:
There are no reviews to evaluate the parameters. So, obviously, the error should be shown.

Expertium:
Oh, that's what you meant. My bad. Yeah, then an error should be shown.

L-M-Sherlock:
OK, this PR will be merged after @dae or @asukaminato0721 reviews it.

Expertium:
> Did Rlz find such cases during the analysis? If there were very few cases, we can skip comparing against the default parameters. This would atleast decrease the time taken for optimization. In addition, this can prevent the confusion arising from a sudden shift to the default parameters.

From a purely "decrease RMSE as much as possible" point of view, this rule is as good as we can get. Any further improvement would be extremely minor. But it's true that we also need to avoid confusion among users. So perhaps skipping comparison with the default parameters would be better, even if it goes against minimizing RMSE at any cost. Initially, I was leaning more towards not skipping it, but now I think it's better to skip it.
I'm typing this out for Dae as well, so that he can understand what's going on.

asukaminato0721 (approved):

Expertium (approved):

src/training.rs
```diff
@@ -490,10 +495,32 @@ mod tests {
...
+        let pretrained_rmse = pretrained_fsrs
+            .evaluate(testset.clone(), |_| true)
+            .unwrap()
+            .rmse_bins;
+        let default_fsrs = FSRS::new(Some(&[])).unwrap();
+        let default_rmse = default_fsrs
+            .evaluate(testset.clone(), |_| true)
+            .unwrap()
+            .rmse_bins;
+        let (best_parameters, best_rmse) = if pretrained_rmse < default_rmse {
```
user1823, line 518: I think that these tests are no longer needed.

Expertium: Good catch

L-M-Sherlock: It is just used for test.

L-M-Sherlock: Any other feedback?

user1823: > Any other feedback?

Nothing more from me.

PR #179, opened 2024-04-17
update burn to 0.13.0

L-M-Sherlock:
There are some breaking changes in burn 0.13.0, but it doesn't affect the API of FSRS.

https://github.com/tracel-ai/burn/releases/tag/v0.13.0

The problem is, we used a custom burn in this branch: https://github.com/open-spaced-repetition/fsrs-rs/tree/fsrs-browser

### DIFF
--- a/src/batch_shuffle.rs
+++ b/src/batch_shuffle.rs
@@ -1,6 +1,7 @@
 use burn::data::{
     dataloader::{
-        batcher::Batcher, BatchStrategy, DataLoader, DataLoaderIterator, FixBatchStrategy, Progress,
+        batcher::DynBatcher, BatchStrategy, DataLoader, DataLoaderIterator, FixBatchStrategy,
+        Progress,
     },
     dataset::Dataset,
 };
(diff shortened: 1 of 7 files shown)


AlexErrant:
I don't think fsrs-browser should block upgrades to fsrs-rs. Please feel free to merge and I'll work through the issues and (eventually) PR to the fsrs-browser branch (and https://github.com/open-spaced-repetition/burn)

L-M-Sherlock:
> Also, Burn recently refactored their autodiff to add a new `std::thread::spawn` invocation

Yeah. We may need to use the `MutexClient`.

By the way, they also removed python dependency and huggingface downloader, so the `mnist.db` was not generated by `cargo run --example guide`: https://github.com/tracel-ai/burn/pull/1283

src/batch_shuffle.rs
asukaminato0721, line 1: here Box instead of Arc, so no Sync now.

asukaminato0721 (approved):

PR #184, opened 2024-04-20
Remove sqrt from count in pretrain

L-M-Sherlock:
The modification in python optimizer: https://github.com/open-spaced-repetition/fsrs-optimizer/pull/106

~Waiting for benchmarking.~

The benchmark result: it reduce RMSE(bins) from 0.059 to 0.058.

For details: https://github.com/open-spaced-repetition/srs-benchmark/commit/ee3a3de0ded934132a97f234ea464bb4de9e01f5

Due to #185 and https://github.com/open-spaced-repetition/fsrs-browser/issues/23, I'm considering downgrade to burn 0.12.1. But I'm afraid it would mess up the commit history and the merge back in the future.

Fine. I decide to wait for the patch of burn.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.6.1"
+version = "0.6.2"
 dependencies = [
  "burn",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fsrs"
-version = "0.6.1"
+version = "0.6.2"
 authors = ["Open Spaced Repetition"]
 categories = ["algorithms", "science"]
 edition = "2021"
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -22,8 +22,8 @@ pub type Parameters = [f32];
 use itertools::izip;
 
 pub static DEFAULT_PARAMETERS: [f32; 17] = [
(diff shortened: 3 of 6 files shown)


dae:
No rush to revert the change - as long as you don't have changes here that you want in Anki, Anki can avoid updating the fsrs-rs version for now.

L-M-Sherlock:
@AlexErrant, this PR could improve the accuracy of FSRS-rs, so I hope we can merge it back to fsrs-browser after we solve https://github.com/open-spaced-repetition/fsrs-browser/issues/23

AlexErrant (approved):
Oh hah just seeing this now, after making the 0.13.1 PRs. Seems good to me, I guess? Not really my area of expertise, but it passes my sanity check.

Just a question, is "0.059 to 0.058" within the realm of noise? Not sure if that's a statistically significant improvement - but this is well beyond my area of knowledge!

L-M-Sherlock:
> Just a question, is "0.059 to 0.058" within the realm of noise? Not sure if that's a statistically significant improvement - but this is well beyond my area of knowledge!

It's statistically significant and it improve the pretraining stage, which is beneficial to learners who have small size of revlog.

AlexErrant:
> beneficial to learners who have small size of revlog

heh, speaking of which: https://news.ycombinator.com/item?id=40176109

I'm not entirely sure what fsrs-rs's minimum training set is, please feel free to correct me!

PR #187, opened 2024-04-27
update fsrs-browser to burn 0.13.1

AlexErrant:
For all merge conflicts, I took main's changes, with the exception of changing the burn version in `Cargo.toml`. I then fixed code in the `fix` commit.


L-M-Sherlock:
https://github.com/open-spaced-repetition/fsrs-rs/actions/runs/8856887500

PR #192, opened 2024-05-16
Fix/calculate current retrievability with FSRS-4.5

L-M-Sherlock:


AlexErrant:
I see these releases 👀 gonna see if I can get auto-releases working with fsrs-browser; might be a bit.

AlexErrant:
Okay it should be working! https://github.com/open-spaced-repetition/fsrs-browser/actions/workflows/sync-with-fsrs-rs.yml

It runs once a day at midnight UTC. You also run it manually (`Run workflow`) if you want a release _right now_ 😸

PR #193, opened 2024-05-21
Replace "weight" with "parameter".rs

Expertium:


src/optimal_retention.rs
```diff
@@ -7,7 +7,7 @@ use ndarray::{s, Array1, Array2, Ix0, Ix1, SliceInfoElem, Zip};
 use ndarray_rand::rand_distr::Distribution;
 use ndarray_rand::RandomExt;
 use rand::{
-    distributions::{Uniform, WeightedIndex},
+    distributions::{Uniform, parameteredIndex},
```
L-M-Sherlock: It's not the weight of FSRS. It's imported from other package. Please revert this change.

L-M-Sherlock (approved):
LGTM

src/optimal_retention.rs
Expertium: My bad, I used auto find and replace

PR #213, opened 2024-08-22
Feat/float next interval

L-M-Sherlock:
Related PR:

- https://github.com/ankitects/anki/pull/3375

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.2.1"
+version = "1.3.0"
 dependencies = [
  "burn",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fsrs"
-version = "1.2.1"
+version = "1.3.0"
 authors = ["Open Spaced Repetition"]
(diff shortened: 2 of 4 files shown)


dae:
These changes look pretty straightforward. I presume we could retain the same behavior on Anki's end, if desired, by rounding the intervals on that side? Is your intention to do that for 1+ day intervals, and treat subday intervals differently if there are no (re)learning steps?

L-M-Sherlock:
Yes. In the corresponding PR, I modified the code to round the interval and make sure it's greater or equal to 1 day. For interval shorter than 0.5 day, if there are no (re)learning steps, it will be converted to seconds.

dae:
FSRS controlling subday intervals is a bit contentious, but we can discuss that in the linked PR - looks like the changes here should be ok to merge?

L-M-Sherlock:
It's OK to merge this PR. But it's nice if we can wait for merging #222, which doesn't need Anki side to change the code. This PR is not back-compatible, if I merge it first, I have to update a bunch of code in Anki side.

L-M-Sherlock:
I recommend merging this PR https://github.com/ankitects/anki/pull/3405 first.

L-M-Sherlock:
@dae, do you approve this PR?

dae (approved):

PR #216, opened 2024-08-23
Fix/simulator crashes if no history

L-M-Sherlock:
source: https://forums.ankiweb.net/t/fsrs-simulator-crashes-anki-if-no-history/48541

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.2.0"
+version = "1.2.1"
 dependencies = [
  "burn",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fsrs"
-version = "1.2.0"
+version = "1.2.1"
 authors = ["Open Spaced Repetition"]
 categories = ["algorithms", "science"]
 edition = "2021"
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,12 +1,13 @@
 use snafu::Snafu;
 
-#[derive(Snafu, Debug)]
+#[derive(Snafu, Debug, PartialEq)]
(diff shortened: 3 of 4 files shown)


asukaminato0721 (approved):

L-M-Sherlock:
@dae should we introduce a new kind of error? Or just return zeros in the simulation result?

dae:
Sorry for the delay. I don't have a strong preference here. One way would be to reuse InvalidInput, and mention in the docstring that deck size is expected to be non-zero (as 0 is nonsensical) - effectively making it the caller's responsibility to check. We could combine that with a tweak on Anki's side so that the size input can't go below 1.

For cases where existing cards > deck size, would it make sense to cap existing cards to the provided value?

dae:
(adding a new error is a valid approach too - I just tend to lump rare/invalid inputs into InvalidInput out of convenience)

L-M-Sherlock:
Sorry for slow response. I'm still in traveling. Here are some my thoughts:

> One way would be to reuse InvalidInput, and mention in the docstring that deck size is expected to be non-zero (as 0 is nonsensical) - effectively making it the caller's responsibility to check.

Sounds good. I will check the code and try to reuse InvalidInput.

> For cases where existing cards > deck size, would it make sense to cap existing cards to the provided value?

Deck size is equal to the number of existing cards plus the number of additional new cards. So existing cards > deck size is invalid because the number of additional new cards should be greater than or equal to zero.

dae:
> Sorry for slow response. I'm still in traveling. Here are some my thoughts:

Not a problem; I hope you're having fun. :-)

> So existing cards > deck size is invalid because the number of additional new cards should be greater than or equal to zero.

Just wondering how much effort we need to be putting into handling non-sensical cases. If we anticipate it to be a common cause of confusion, a separate message to the user about how it doesn't make sense could work. But if this is going to be pretty rare, a simpler solution like clamping the value or returning 0s might be fine too.

PR #217, opened 2024-08-29
Check parameters are valid numbers

dae:
Follow-up to #214


src/model.rs
```diff
@@ -256,6 +256,9 @@ pub(crate) fn check_and_fill_parameters(parameters: &Parameters) -> Result<Vec<f
         19 => parameters.to_vec(),
         _ => return Err(FSRSError::InvalidParameters),
     };
+    if parameters.iter().any(|&w| !w.is_finite()) {
```
asukaminato0721, line 259:
```rs
w.is_infinite()
```

?

asukaminato0721: never mind, I forgot nan

asukaminato0721 (approved):

PR #219, opened 2024-09-07
Change the R_MAX of CMRR.rs

Expertium:
https://github.com/open-spaced-repetition/fsrs4anki/issues/686#issuecomment-2335482240
It appears that our previous estimates of optimal retention were underestimates. I think it makes sense to expand the range.
image
Note that here it says "workload", but it's not just workload, it's workload/knowledge

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -56,7 +56,7 @@ impl From<Column> for SliceInfoElem {
 }
 
 const R_MIN: f32 = 0.75;
-const R_MAX: f32 = 0.95;
+const R_MAX: f32 = 0.98;
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct SimulatorConfig {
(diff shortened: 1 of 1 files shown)


L-M-Sherlock (approved):
LGTM

user1823:
I think that DR above 0.96 doesn't make much sense because the returns are diminishing.

image

In this image, see the drastic increase in slope above DR = 0.96. (There are occasional high slopes before DR = 0.96 too but after DR = 0.96, most of the values are too high.)

L-M-Sherlock:
The data is based on default parameters.

user1823:
Well, it is OK if you think that some people can have such a high value of MRR. But, I am not really convinced. Nonetheless, let's forget about this for now. We can always decrease the value if someone complains.

Expertium:
We can change it slightly to 0.97

Expertium:
@L-M-Sherlock do you think we should revert this? It seems like the idea that previous values were underestimates was based on the wrong data

PR #222, opened 2024-09-10
Feat/fine-tune the sample size for CMRR

L-M-Sherlock:


dae:
I can't check your math/logic, but the Rust looks fine :-)

asukaminato0721 (approved):

Expertium:
A bit late, but I recommend adding an assertion that the sample size is always between 45 and 4, and never >45 or <4.

L-M-Sherlock:
I have tested it. When `learn_span` is 31, the sample size is 45. When `learn_span` is 364, the sample size is 4.

PR #229, opened 2024-10-04
R_MIN = 0.70

Expertium:
I have tried the most recent beta: https://github.com/ankitects/anki/releases/tag/24.10beta2

CMRR values have gone down for every single one of my presets, meaning that the old values were overestimates. Out of 7 presets that I tried CMRR on, I got 0.75 for three of them, and 0.90 for only one of them.

@user1823, @brishtibheja I'd like you to optimize your parameters and try CMRR in Anki 24.10 beta 2. If your CMRR values are also only going down and not up, then it would imply that the range needs to be extended towards lower values.

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -55,7 +55,7 @@ impl From<Column> for SliceInfoElem {
     }
 }
 
-const R_MIN: f32 = 0.75;
+const R_MIN: f32 = 0.70;
 const R_MAX: f32 = 0.95;
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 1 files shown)


user1823:
For me, MRR decreased from 0.87 to 0.75.

But, before decreasing the range, we should try to find the reason why the value is going down. Is there any logical explanation? Or do we have a bug?

Expertium:
Good question.
Here's the old workload/retention graph:
image

New one:
image

It's hard to compare them because the old one doesn't have text that says "minimum workload x(value)", but basically in the old one it's x1, x2, x3, x4. The new graph suggests that previously we overestimated how quickly workload increases as retention increases. The new graph is less steep. However, this doesn't automatically mean that the minimum has shifted, unless I'm bad at math. So idk.

L-M-Sherlock (approved):
LGTM

Expertium:
We should've probably discussed it a bit more...

PR #238, opened 2024-10-16
Fix/sample_size should be continuous over learn_span

Expertium:
It should be 8, not 16. Originally, the maximum sample size was 4. Since you doubled default_sample_size, this should also double.


Expertium:
I'm curious to know whether it's still faster than before even with this big change in the sample size

L-M-Sherlock (approved):
LGTM

PR #239, opened 2024-10-20
Feat/linear damping

L-M-Sherlock:
close https://github.com/open-spaced-repetition/fsrs4anki/issues/697

@dae, I make a minor change in the model of FSRS-5. I prefer not to update the algorithm version (e.g. FSRS-5.1, FSRS-5.5 or something else) because Anki has stored FSRS-5 parameters separately and FSRS-5 hasn't been released in stable version of Anki. What do you think of?


dae:
My assumption with fsrs_params_5 is that we can continue to use it for future FSRS 5.x releases. We haven't had a stable release with 5 yet, so I think this change is fine - we can suggest users optimise again after updating to the next beta/RC.

L-M-Sherlock:
> we can suggest users optimise again after updating to the next beta/RC

Yes, it's exactly what I wanted to say.

asukaminato0721 (approved):

dae (approved):

PR #253, opened 2024-11-19
feat: derive Hash for FSRSReview and FSRSItem

sineptic:


L-M-Sherlock:
Thank you for your contribution! Could you please provide more details on why you decided to derive the `Hash` trait for `FSRSReview` and `FSRSItem`? Understanding the motivation behind this change will help us in the review process.

sineptic:
Excuse me please. I feeling so guilty, I don't think before trying to contribute.

My task manager type has `tasks_to_complete` and `tasks_pool`. I wrote function `find_tasks_to_recall`, that (I think) find tasks to recall in fastest way, but only approach for this is implement `Cmp`(compare time to recall) for tasks wrapper(to use in BTreeSet and stop when first not to complete found). I quickly realize, that comparing time with default parameters not work for me, so `BTreeSet` -> `HashSet` and scan all for tasks to complete, but I don't think that `Vec` could replace `HashSet`

Issue #256, opened 2024-12-08
[TODO] support sample weights when training


L-M-Sherlock:
It will make these possible:
- give more weights on post-lapse reviews to improve the accuracy on post-lapse stability.
- give more weights on recent reviews (Time Decay) to fit the learner's current memory pattern and review habit.

fsrs-optimizer has supported it: https://github.com/open-spaced-repetition/fsrs-optimizer/pull/152

brishtibheja:
This is probably going to improve things for end-user but when we update our Anki and optimise, this should result in worse log-loss in which Anki will just retain the previous parameters instead of updating them. Wdyt about that?

L-M-Sherlock:
> This is probably going to improve things for end-user but when we update our Anki and optimise, this should result in worse log-loss in which Anki will just retain the previous parameters instead of updating them. Wdyt about that?

Maybe we need apply the weights in the evaluation, too.

Expertium:
I've said this before - if we apply weights during both evaluation and training, then loss values will effectively become arbitrary since they depend on the weighting function.

PR #258, opened 2024-12-16
Feat/option enable_short_term in training

### LINKED ISSUE #255: [TODO] Option to disable `stability_short_term`
It will freeze `w[17]` and `w[18]` to zeros when training.

It would partially solve:
- https://github.com/open-spaced-repetition/fsrs4anki/issues/708

L-M-Sherlock:
close #255 

The API is changed in this PR, but I don't want to release a version because #256 also requires a major change.

What's the best practice for this case?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.4.4"
+version = "1.5.0"
 dependencies = [
  "burn",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.5.0"
+version = "2.0.0"
 dependencies = [
  "burn",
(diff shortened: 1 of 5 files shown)


asukaminato0721:
in py, we use something like default parm so the api doesn't break.

```py
def f(val = true)
f()
```

but since rust doesn't have it

one way is to separate the method

```rs
fn f( val: bool )
```

to

```rs
fn f() // by default true
fn f_false()
```

so the api doesn't break.

L-M-Sherlock:
@asukaminato0721 cc f0f2542

asukaminato0721 (approved):

L-M-Sherlock:
@dae, I find it's really hard to solve the problems mentioned in https://github.com/open-spaced-repetition/fsrs4anki/issues/708, so I decided to provide an option to freeze the stability in FSRS-5 if the user has trouble with the same-day reviews. What do you think of?

L-M-Sherlock:
Because #267 will introduce a breaking change, I decide not to use partial function.

PR #260, opened 2024-12-19
Feat/support recency weighting

### LINKED ISSUE #256: [TODO] support sample weights when training

L-M-Sherlock:
close #256

same to:
- https://github.com/open-spaced-repetition/fsrs-optimizer/pull/152

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -6,8 +6,8 @@ use burn::nn::loss::Reduction;
 use burn::tensor::{Data, Shape, Tensor};
 use burn::{data::dataloader::batcher::Batcher, tensor::backend::Backend};
 
-use crate::dataset::FSRSBatch;
 use crate::dataset::FSRSBatcher;
+use crate::dataset::{simple_weighted_fsrs_items, FSRSBatch};
 use crate::error::Result;
 use crate::model::Model;
 use crate::training::BCELoss;
@@ -210,9 +210,11 @@ impl<B: Backend> FSRS<B> {
         if items.is_empty() {
             return Err(FSRSError::NotEnoughData);
         }
+        let items = simple_weighted_fsrs_items(items);
         let batcher = FSRSBatcher::new(self.device());
         let mut all_retention = vec![];
         let mut all_labels = vec![];
+        let mut all_weights = vec![];
         let mut progress_info = ItemProgress {
             current: 0,
             total: items.len(),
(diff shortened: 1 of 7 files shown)


asukaminato0721 (approved):

src/inference.rs
```diff
@@ -210,29 +219,33 @@ impl<B: Backend> FSRS<B> {
         if items.is_empty() {
             return Err(FSRSError::NotEnoughData);
         }
+        let weighted_items = recency_weighted_fsrs_items(items);
```
user1823, line 222: @L-M-Sherlock, shouldn't this be `constant_weighted_fsrs_items`?

In https://github.com/open-spaced-repetition/fsrs-optimizer/pull/152#issuecomment-2548261225, you said that evaluation doesn't use weights.

In addition to what @Expertium said, adding weights to the evaluation makes comparing RMSE in Anki 24.11 and 25.01 difficult.

I noticed this because Anki 24.11 shows `Log loss: 0.2551, RMSE(bins): 2.91%. 64,368 reviews` while Anki 25.01 shows `Log loss: 0.2891, RMSE(bins): 3.18%. 64,368 reviews` with the same parameters.

L-M-Sherlock: It should be `recency_weighted_fsrs_items` otherwise Anki will prefer the old parameters.

user1823: Well, I have changed my mind. Recency weighting makes training give more importance to recent reviews, which are those that the users care the most about. 

If evaluation doesn't apply recency weighting, the more optimal parameters will produce worse metrics because the older reviews will have higher loss. So, Anki won't accept these new parameters.

But, we will need to inform users that this increased RMSE is due to a change in the evaluation method and this increase doesn't mean that FSRS has become worse for them.

L-M-Sherlock: I doubt most users will check the RMSE after they update Anki. We just need to inform a small group of users if necessary.

Btw, RMSE(bins) cannot tell us much useful information. If we want to diagnose the algorithm, some calibration graphs in the section 4.2 is necessary: https://github.com/open-spaced-repetition/fsrs4anki/blob/main/fsrs4anki_optimizer.ipynb

PR #235, opened 2024-10-10
Make simulate iterate by card instead of by day.

Luc-Mcgrady:
Simulating the reviews iterating card by card instead of day by day speeds up the calculation (especially with a long learn_span) and simplifies the code.
Also fixes the fact that there's a drop-off near the start.

I hope that this functions in a way pretty much exactly like the day by day version, I'll leave the tests failing so you can see exactly how the results changed.

```js
thread 'optimal_retention::tests::simulate_with_learn_review_limit' panicked at src/optimal_retention.rs:909:9:
assertion `left == right` failed
  left: [28, 21, 40, 80, 72, 81, 86, 102, 72, 93, 105, 115, 120, 118, 124, 126, 137, 136, 158, 150, 159, 171, 156, 170, 183, 158, 195, 173, 194, 178]
 right: [0, 15, 16, 27, 69, 66, 81, 83, 75, 85, 90, 102, 107, 105, 122, 120, 141, 117, 132, 153, 153, 154, 152, 152, 149, 157, 166, 165, 184, 174]
```
```js
thread 'optimal_retention::tests::simulator' panicked at src/optimal_retention.rs:864:9:
assertion `left == right` failed
  left: 7778.458
 right: 6521.068
```
```js
thread 'optimal_retention::tests::optimal_retention' panicked at src/optimal_retention.rs:976:9:
assertion `left == right` failed
  left: 0.86305356
 right: 0.8277919
``` 
(/\ I've probably screwed up somewhere)
```js
thread 'optimal_retention::tests::optimal_retention_with_old_parameters' panicked at src/optimal_retention.rs:996:9:
assertion `left == right` failed
  left: 0.8716841
 right: 0.85450846
```
Please note that my results are left shifted one value compared to before because "today" will have a very small cost if the user has already done their reviews. (I think).

And here's how the graphs look with my decks:

### This pr
new
### Before
(Starting drop-off lasts more than one day so not just the fact reviews are done, if I had to guess the first review of a card isn't registered for some reason)
old
### This pr
image
### Before
image
### This pr
image
### Before
image

PR #262, opened 2024-12-21
Fix/check review_limit only if is_learn is false

L-M-Sherlock:
fix https://github.com/ankitects/anki/issues/3654

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.4.6"
+version = "1.4.7"
 dependencies = [
  "burn",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fsrs"
-version = "1.4.6"
+version = "1.4.7"
 authors = ["Open Spaced Repetition"]
 categories = ["algorithms", "science"]
 edition = "2021"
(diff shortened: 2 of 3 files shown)


Luc-Mcgrady (approved):
That issue is talking about this option
image

which defaults to off though in the simulator defaults to on.

image

To fix it you would need to add that to the simulator config.
something like.

```rs
let review_learn_count = learn_cnt_per_day[day_index] * (config.new_cards_affect_review_limit ? 1 : 0) 
if (!is_learn && review_cnt_per_day[day_index] + review_learn_count + 1 > review_limit)
    || (is_learn && learn_cnt_per_day[day_index] + 1 > learn_limit)
    || (cost_per_day[day_index] + fail_cost > max_cost_perday)
```

I originally made it default to off in the simulator:
https://github.com/open-spaced-repetition/fsrs-rs/pull/235/files/a79f4221e3652aedd940720b77ad0774407da736..cbe8f9d98f75189445d90435ec7339e900b01fa0#r1798205551
 
This is still a bug though I'm pretty sure :smile:

L-M-Sherlock:
OK, so we should make a breaking change?

Luc-Mcgrady:
This does actually fix this issue nvm.

I'm not sure its worth the breaking change, maybe if another breaking change happens we can include it then.

L-M-Sherlock:
I tend to merge this PR and keep the breaking change in the future. What do you think of?

Luc-Mcgrady:
Sounds good

PR #267, opened 2024-12-30
Feat/Simulator learn affects review limit option

Luc-Mcgrady:
What I was talking about here: https://github.com/open-spaced-repetition/fsrs-rs/pull/262#pullrequestreview-2518661668

Implemented in Anki: https://github.com/Luc-Mcgrady/anki/tree/simulator-new-ignores-review-limit

Orange = true
Blue = false

Breaking but small change so I'd recommend waiting for something more important before merging

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.4.9"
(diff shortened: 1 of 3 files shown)


L-M-Sherlock:
It introduced a breaking change in `simulate` API. I consider to include it and #258 in v2.0.0.

PR #270, opened 2025-01-10
Feat/L2 Regularization

L-M-Sherlock:
related PR:
- https://github.com/open-spaced-repetition/fsrs-optimizer/pull/157

### DIFF
--- a/.github/workflows/check.sh
+++ b/.github/workflows/check.sh
@@ -13,5 +13,5 @@ unzip *.zip
 
 RUSTDOCFLAGS="-D warnings" cargo doc --release
 
-cargo install cargo-llvm-cov --locked
+cargo install cargo-llvm-cov@0.6.15 --locked
 SKIP_TRAINING=1 cargo llvm-cov --release
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "2.0.1"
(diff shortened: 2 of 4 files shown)


Expertium:
I'm currently trying out a different regularization method (well, it wasn't really made _for_ regularization, but it kind of acts like a regularizer and improves generalization). If it works well, we can ditch L2 which barely improves the metrics. So don't merge this just yet.

Expertium:
The results ended up being underwhelming.
I implemented AutoClip for gradients. Relevant repo: https://github.com/pseeth/autoclip. The code is very simple, I implemented it in other.py in just 16 lines.
The key idea is simple: keep track of observed gradient norms, calculate the p-th percentile (`clip_value = np.percentile(grad_history, clip_percentile)`) of them, then if next time the gradient norm exceeds that value, use `torch.nn.utils.clip_grad_norm_(self.model.parameters(), clip_value)`. This is kind of like outlier filtering, but with gradients.
From their paper: 
image

The authors recommend using p=10 aka 10th percentile, which is extremely aggressive, but perhaps suitable for large neural nets. 
image
I tried different values of `clip_percentile` on FSRS-5, and sadly it barely improves the metrics.
So let's use L2 regularization.

L-M-Sherlock:
Wow, L2 regularization has very significant impact on the distribution of w[12], w[14], w[15], w[16]

image

image

image

image

asukaminato0721 (approved):

PR #283, opened 2025-02-18
Refactor/provide a hook for post scheduling adjustments in simulator configuration

L-M-Sherlock:
This PR introduces several significant changes to the simulator configuration and interval scheduling mechanism:

### Major Changes
1. Replaced `ndarray::Array1` with standard `Vec` for simulation result storage
2. Added customizable post-scheduling functionality through a new `PostSchedulingFn` type
3. Removed built-in fuzzing logic in favor of configurable post-scheduling
4. Added manual implementations for `PartialEq` and `Debug` for `SimulatorConfig`

### Implementation Details
- Introduced a new type alias `PostSchedulingFn` for post-scheduling operations
- Added `post_scheduling_fn` field to `SimulatorConfig` to support custom interval adjustments
- Simplified interval calculation by removing the built-in fuzzing mechanism
- Updated test cases to reflect new behavior with modified expected values

### Breaking Changes
- Removed `fuzz_bounds`, `constrained_fuzz_bounds`, and related fuzzing functions
- Changed simulation result types from `Array1<T>` to `Vec<T>`
- Modified interval scheduling behavior which may affect simulation results

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -5,7 +5,6 @@ use crate::parameter_clipper::clip_parameters;
 use crate::FSRS;
 use burn::tensor::backend::Backend;
 use itertools::{izip, Itertools};
-use ndarray::{s, Array1};
 use ndarray_rand::rand_distr::Distribution;
 use priority_queue::PriorityQueue;
 use rand::Rng;
@@ -17,10 +16,10 @@ use std::collections::HashMap;
 
 #[derive(Debug)]
 pub struct SimulationResult {
-    pub memorized_cnt_per_day: Array1<f32>,
-    pub review_cnt_per_day: Array1<usize>,
-    pub learn_cnt_per_day: Array1<usize>,
-    pub cost_per_day: Array1<f32>,
+    pub memorized_cnt_per_day: Vec<f32>,
+    pub review_cnt_per_day: Vec<usize>,
+    pub learn_cnt_per_day: Vec<usize>,
+    pub cost_per_day: Vec<f32>,
 }
 
 trait Round {
@@ -36,7 +35,6 @@ impl Round for f32 {
 const R_MIN: f32 = 0.70;
 const R_MAX: f32 = 0.95;
 
-#[derive(Debug, Clone, PartialEq)]
 pub struct SimulatorConfig {
     pub deck_size: usize,
     pub learn_span: usize,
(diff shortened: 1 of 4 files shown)


src/optimal_retention.rs
```diff
@@ -36,7 +35,11 @@ impl Round for f32 {
 const R_MIN: f32 = 0.70;
 const R_MAX: f32 = 0.95;
 
-#[derive(Debug, Clone, PartialEq)]
+/// Function type for post scheduling operations that takes interval, maximum interval,
+/// current day index, due counts per day, and a random number generator,
+/// and returns a new interval.
+pub type PostSchedulingFn = dyn Fn(f32, f32, usize, Vec<usize>, &mut StdRng) -> f32 + Send + Sync;
```
asukaminato0721, line 41: is the following method ok?

```rs
struct Wrap(dyn Fn(f32, f32, usize, Vec<usize>, &mut StdRng) -> f32 + Send + Sync)
```

something like this

then `impl PartialEq` just return true (

`impl Debug` return `"<function>"`

asukaminato0721: create a new type, bypass the orphan rule.

src/optimal_retention.rs
```diff
@@ -54,6 +57,55 @@ pub struct SimulatorConfig {
...
+            && self.first_rating_prob == other.first_rating_prob
+            && self.review_rating_prob == other.review_rating_prob
+            && self.first_rating_offsets == other.first_rating_offsets
+            && self.first_session_lens == other.first_session_lens
+            && self.forget_rating_offset == other.forget_rating_offset
+            && self.forget_session_len == other.forget_session_len
+            && self.loss_aversion == other.loss_aversion
+            && self.learn_limit == other.learn_limit
+            && self.review_limit == other.review_limit
+            && self.new_cards_ignore_review_limit == other.new_cards_ignore_review_limit
```
asukaminato0721, line 80: so we don't need to `==` all the fields.

src/optimal_retention.rs
```diff
@@ -478,6 +390,20 @@ pub fn simulate(
             }
         }
 
+        let mut ivl = next_interval(card.stability, desired_retention)
+            .round()
+            .clamp(1.0, config.max_ivl);
+
+        if let Some(ref post_scheduling_fn) = config.post_scheduling_fn {
```
asukaminato0721, line 397: and here just `config.post_scheduling_fn.0`

src/optimal_retention.rs
L-M-Sherlock: see ec0b264

src/optimal_retention.rs
```diff
@@ -36,7 +35,27 @@ impl Round for f32 {
...
+/// Function type for post scheduling operations that takes interval, maximum interval,
+/// current day index, due counts per day, and a random number generator,
+/// and returns a new interval.
+pub type PostSchedulingFnInner =
+    dyn Fn(f32, f32, usize, Vec<usize>, &mut StdRng) -> f32 + Send + Sync;
+
+pub struct PostSchedulingFn(pub Box<PostSchedulingFnInner>);
+
+impl PartialEq for PostSchedulingFn {
+    fn eq(&self, _other: &Self) -> bool {
```
asukaminato0721, line 47:
```suggestion
    fn eq(&self, _: &Self) -> bool {
```

asukaminato0721 (approved):

PR #284, opened 2025-02-18
Feat/support review priority

L-M-Sherlock:


Luc-Mcgrady (approved):
LGTM (Wait hold on why do the tests fail?)

src/optimal_retention.rs
```diff
@@ -378,7 +417,10 @@ pub fn simulate(
             due_cnt_per_day[card.due as usize] += 1;
         }
 
-        card_priorities.change_priority(&card_index, card_priority(card, false));
+        card_priorities.change_priority(
+            &card_index,
+            card_priority(card, is_learn, config.review_priority_fn.as_ref()),
```
Luc-Mcgrady, line 422:
```suggestion
            card_priority(card, false, config.review_priority_fn.as_ref()),
```

Luc-Mcgrady (changes requested):

src/optimal_retention.rs
Luc-Mcgrady: Yep, now you've merged the other pull request changing this fixed the tests for me :+1:

Luc-Mcgrady (approved):

src/optimal_retention.rs
```diff
@@ -227,13 +247,29 @@ pub fn simulate(
...
+    fn card_priority(
+        card: &Card,
+        learn: bool,
+        review_priority_fn: Option<&ReviewPriorityFn>,
+    ) -> Reverse<(i32, bool, i32)> {
+        let priority = if let Some(priority_fn) = review_priority_fn {
+            priority_fn.0(card)
+        } else {
+            (card.difficulty * 100.0) as i32
+        };
```
asukaminato0721, line 259:
```suggestion
        let priority = review_priority_fn.map_or((card.difficulty * 100.0) as i32, |priority_fn| priority_fn.0(card));
```

can this work?

asukaminato0721 (approved):

src/optimal_retention.rs
Luc-Mcgrady: Could this let be moved outside of the priority function and passed as the argument instead since its value doesn't change? Or would the optimizer do this anyway?

PR #286, opened 2025-02-19
Refactor/Use default_review_priority instead of map_or

Luc-Mcgrady:
Moving the "if" outside of the review_priority function is neater IMO

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -247,14 +247,21 @@ pub fn simulate(
 
     let mut card_priorities = PriorityQueue::new();
 
+    fn default_review_priority(card: &Card) -> i32 {
+        (card.difficulty * 100.0) as i32
+    }
+    let default_review_priority = ReviewPriorityFn(Box::new(default_review_priority));
+    let review_priority_fn = config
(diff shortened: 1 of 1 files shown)


src/optimal_retention.rs
```diff
@@ -247,14 +247,21 @@ pub fn simulate(
 
     let mut card_priorities = PriorityQueue::new();
 
+    fn default_review_priority(card: &Card) -> i32 {
+        (card.difficulty * 100.0) as i32
+    }
+    let default_review_priority = ReviewPriorityFn(Box::new(default_review_priority));
```
asukaminato0721, line 253: why not `impl Default for ReviewPriorityFn`?

then `unwrap_or_default`

src/optimal_retention.rs
```diff
@@ -247,14 +247,21 @@ pub fn simulate(
...
+    let review_priority_fn = config
+        .review_priority_fn
+        .as_ref()
+        .unwrap_or(&default_review_priority);
```
asukaminato0721, line 257:
```suggestion
        .unwrap_or_default();
```

PR #292, opened 2025-02-21
some fix?

### LINKED ISSUE #291: check whether the function sig can be simplify
miss push the commit to the `main`, need hotfix it.

asukaminato0721:
close #291 

Idk whether the inner type need to be pub?

and use &[ ] since Vec is not needed.

<https://github.com/rust-lang/rust-clippy/issues/9299> lol.


L-M-Sherlock (approved):
LGTMLGTM

PR #293, opened 2025-02-23
Feat/Simulator Suspend After Lapse Count

Luc-Mcgrady:
Tested with: https://github.com/Luc-Mcgrady/anki/tree/suspend-after-lapse-count

Limits for examples:
(1: no limit, 2: 14, 3: 7, 4: 100)
image
image

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -97,6 +97,7 @@ pub struct SimulatorConfig {
     pub learn_limit: usize,
     pub review_limit: usize,
     pub new_cards_ignore_review_limit: bool,
+    pub suspend_after_lapses: Option<u32>,
     pub post_scheduling_fn: Option<PostSchedulingFn>,
     pub review_priority_fn: Option<ReviewPriorityFn>,
 }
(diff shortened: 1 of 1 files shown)


src/optimal_retention.rs
```diff
@@ -366,6 +371,8 @@ pub fn simulate(
             // Create 'forget' mask
             let forget = !rng.gen_bool(retrievability as f64);
 
+            card.lapses += if forget { 1 } else { 0 };
```
asukaminato0721, line 374: try :)

```suggestion
            card.lapses += forget as u32;
```

src/optimal_retention.rs
```diff
@@ -120,6 +121,7 @@ impl Default for SimulatorConfig {
             learn_limit: usize::MAX,
             review_limit: usize::MAX,
             new_cards_ignore_review_limit: true,
+            suspend_after_lapses: None,
             post_scheduling_fn: None,
             review_priority_fn: None,
```
asukaminato0721, line 126: maybe try

```suggestion
            ..Default::default()
```

src/optimal_retention.rs
Luc-Mcgrady: nice! :)

src/optimal_retention.rs
Luc-Mcgrady: This causes a stack overflow, it just calls ```fn default() -> Self {``` repeatedly.

L-M-Sherlock (approved):
LGTM

PR #295, opened 2025-03-03
Fix/Export SimulationResult

Luc-Mcgrady:
I need this because I'm trying to add python bindings :pensive:.
https://github.com/open-spaced-repetition/fsrs-rs-python/pull/23#issue-2891108668


L-M-Sherlock (approved):
LGTM

PR #297, opened 2025-03-05
Feat/consider num_relearning_steps when clamp params

L-M-Sherlock:
The problem this PR wants to solve is https://github.com/open-spaced-repetition/fsrs-rs/pull/297#issuecomment-2700243555

@dae, I add an optional arg for `compute_parameters` and `benchmark`, is this design OK? If the API is stable, I will release v3.0.0 of fsrs-rs to crate.

@Luc-Mcgrady, would you have more breaking changes?


Expertium:
To elaborate on why we need this fix: some users, especially Ishmaelite Caesar from Discord, are having significantly worse RMSE after this fix: https://github.com/ankitects/anki/pull/3742

Setting the last two parameters to 0 makes metrics worse.
image-6
image-5

The idea is to set them to some small non-zero values to improve the metrics without reverting the fix of the original issue of post-lapse intervals being longer than pre-lapse intervals.

dae:
Because the number of input parameters is growing, you may want to consider creating a ComputeParametersInput structure and putting all the args in that instead. If we keep adding params, clippy will eventually suggest doing that, but the threshold for its warning is probably 1-2 more (and thus the current approach is also borderline ok). Your call :-)

src/training.rs
```diff
@@ -329,12 +354,14 @@ impl<B: Backend> FSRS<B> {
...
-        num_relearning_steps: Option<usize>,
-    ) -> Vec<f32> {
+    pub fn benchmark(&self, input: ComputeParametersInput) -> Vec<f32> {
+        let ComputeParametersInput {
+            train_set,
+            enable_short_term,
+            num_relearning_steps,
+            ..
+        } = input;
+
```
asukaminato0721, line 364:
```suggestion
    pub fn benchmark(&self, ComputeParametersInput {
            train_set,
            enable_short_term,
            num_relearning_steps,
            ..
        } : ComputeParametersInput) -> Vec<f32> {
```

I guess this will be shorter.

other places as well.

src/training.rs
```diff
@@ -228,15 +228,40 @@ pub fn calculate_average_recall(items: &[FSRSItem]) -> f32 {
...
-        enable_short_term: bool,
-        num_relearning_steps: Option<usize>,
-    ) -> Result<Vec<f32>> {
+    pub fn compute_parameters(&self, input: ComputeParametersInput) -> Result<Vec<f32>> {
+        let ComputeParametersInput {
+            train_set,
+            progress,
+            enable_short_term,
+            num_relearning_steps,
+        } = input;
```
asukaminato0721, line 263: here as well.

asukaminato0721 (approved):

PR #298, opened 2025-03-09
Feat: Pass card instead of ivl to post scheduling function.

Luc-Mcgrady:
This future proofs the simulator in case any ideas like the deadline:
https://github.com/Luc-Mcgrady/anki/commit/fbff2f8318f5d995fc03b1a9cca5a18ab74c2d35
https://forums.ankiweb.net/t/deadline-exam-date-feature/56675/3?u=a_blokee
come to fruition.

Could be a better idea just to include stability as an argument? I just don't want 3.0.0 to release and be locked out of simulating this for the foreseeable future :laughing:.

Also exports `next_interval` from `inference`.


L-M-Sherlock (approved):
LGTM

PR #306, opened 2025-03-16
Feat/Simulator Correct per day

Luc-Mcgrady:
> @Luc-Mcgrady, do you have more breaking changes?

Thanks for asking!
This will ~~probably~~ have to be my last.

I made this to try and help with the load balancer debate. (reviews/day here = average retention)
https://discord.com/channels/368267295601983490/1282005522513530952/1349124940787486813

https://github.com/Luc-Mcgrady/anki/tree/correct-over-time
image
image
image

I didn't pr this because it seems to act weird when adding new cards:
image

Maybe just ignore new cards for this?


L-M-Sherlock:
I my opinion, you shouldn't count learn card in correct rate, because:

https://github.com/open-spaced-repetition/fsrs-rs/blob/6bf567689ef11fa3eaa297fff44f95fbfaa72afd/src/optimal_retention.rs#L355

It's not based on the desired retention.

L-M-Sherlock:
> How are there 5 reviews on the second day with existing cards:

Because they are review cards and their due is 1. Any problem here?

Luc-Mcgrady:
Ohhhh I forgot I added 5 😅

Luc-Mcgrady:
If there is a problem it can be fixed later with a non breaking change. I suspect this is more likely a problem with my Anki implementation.

L-M-Sherlock (approved):
LGTM

Issue #309, opened 2025-03-25
dependabot config?


mikehardy:
Was updating deps in Anki-Android-Backend last couple days, playing with dependabot for the cargo ecosystem and noticed that the transitives coming from fsrs are stale

You might something like this, it's similar to the options I usually run with to keep the noise down, value-per-pull up while keeping major vers separate since those are the frequently problematic ones requiring forward-port touchups

`.github/dependabot.yml`

```yaml
version: 2
updates:
- package-ecosystem: cargo
  directory: "/"
  schedule:
    interval: weekly
  groups:
    non-breaking:
      update-types:
        - "minor"
        - "patch"
  open-pull-requests-limit: 10
```

Obviously close if you are uninterested, just that I'd drop it in case it was useful

L-M-Sherlock:
We are making breaking changes recently, so I haven't released the latest commit to crate. What's the use of this bot? Does it fit the current case?

PR #310, opened 2025-03-26
Create dependabot.yml

asukaminato0721:
fix https://github.com/open-spaced-repetition/fsrs-rs/issues/309

well, who cares. Just randomly pick one.

https://docs.renovatebot.com/bot-comparison/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: weekly
+  groups:
+    non-breaking:
+      update-types:
+        - "minor"
+        - "patch"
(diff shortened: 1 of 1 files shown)


mikehardy:
FWIW - I like dependabot for ecosystems it is good at handling, and after testing how it works with cargo, it seems okay for that. It is built in to github, so that's nice if you are already on github.

Renovate is not bad though, and works well in cases where dependabot doesn't support an ecosystem well.

Either way, after the first flurry of dependencies, it should settle down - it looks like you are using my options from Anki-Android-Backend so it shouldn't be too noisy. If I change the options in Anki-Android-Backend for some reason I'll try to remember to post a PR here to suggest a sync and why. Goal will always be to increase signal-to-noise while keeping things up to date with near zero effort

Cheers

PR #312, opened 2025-03-27
add test for good-again loop during the same day

L-M-Sherlock:
This test reproduced the case 3 mentioned in https://github.com/open-spaced-repetition/fsrs4anki/issues/708:

> 3\. If the user is in a Good - Again - Good - Again loop (during the same day), stability will either explode or fall to near 0. This is possible if his learning steps are, for example, `10m 30m`.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -642,6 +642,26 @@ mod tests {
 
     #[test]
     #[ignore = "just for exploration"]
+    fn good_again_loop_during_the_same_day() -> Result<()> {
+        let fsrs = FSRS::new(Some(&DEFAULT_PARAMETERS))?;
+        let mut state = MemoryState {
+            stability: 1.0,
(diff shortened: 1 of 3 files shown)


src/inference.rs
```diff
@@ -640,6 +640,26 @@ mod tests {
...
+    #[test]
+    #[ignore = "just for exploration"]
+    fn good_again_loop_during_the_same_day() -> Result<()> {
+        let fsrs = FSRS::new(Some(&DEFAULT_PARAMETERS))?;
+        let mut state = MemoryState {
+            stability: 1.0,
+            difficulty: 5.0,
+        };
+
+        let mut stability = Vec::new();
```
asukaminato0721, line 652:
```suggestion
        let mut stability = Vec::with_capacity(10);
```

asukaminato0721 (approved):

PR #322, opened 2025-04-26
Fix/Missing-simulator-decay-for-fsrs5

Luc-Mcgrady:
:+1: 

related: https://forums.ankiweb.net/t/anki-25-05-beta-1/59710/15?u=a_blokee

Edit: How is this error even possible?

https://github.com/open-spaced-repetition/fsrs-rs/blob/3e2f0b423fed194d238cdcb55c1baccd0eca63f0/src/model.rs#L294-L298

Is a deeper fix than this needed?

Edit2: The bug is the weights for the priority functions don't have `check_and_fill_parameters` applied to them. I think this is the best fix after all.

https://github.com/ankitects/anki/blob/2406830ff949668403c60f52d56201f456e5a2d9/rslib/src/scheduler/fsrs/simulator.rs#L96-L101

### DIFF
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -1,7 +1,7 @@
-use crate::{FSRS, FSRS5_DEFAULT_DECAY};
 use crate::error::{FSRSError, Result};
 use crate::inference::{ItemProgress, Parameters, S_MAX, S_MIN};
 use crate::model::check_and_fill_parameters;
+use crate::{FSRS, FSRS5_DEFAULT_DECAY};
 use burn::tensor::backend::Backend;
 use itertools::{Itertools, izip};
 use ndarray_rand::rand::distributions::WeightedIndex;
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -1,4 +1,4 @@
-use crate::FSRS;
+use crate::{FSRS, FSRS5_DEFAULT_DECAY};
 use crate::error::{FSRSError, Result};
 use crate::inference::{ItemProgress, Parameters, S_MAX, S_MIN};
 use crate::model::check_and_fill_parameters;
@@ -232,13 +232,13 @@ fn mean_reversion(w: &[f32], init: f32, current: f32) -> f32 {
 }
 
 fn power_forgetting_curve(w: &[f32], t: f32, s: f32) -> f32 {
-    let decay = -w[20];
+    let decay = -w.get(20).unwrap_or(&FSRS5_DEFAULT_DECAY);
     let factor = 0.9f32.powf(1.0 / decay) - 1.0;
     (t / s).mul_add(factor, 1.0).powf(decay)
 }
 
 fn next_interval(w: &[f32], stability: f32, desired_retention: f32) -> f32 {
-    let decay = -w[20];
+    let decay = -w.get(20).unwrap_or(&FSRS5_DEFAULT_DECAY);
     let factor = 0.9f32.powf(1.0 / decay) - 1.0;
     stability / factor * (desired_retention.powf(1.0 / decay) - 1.0)
 }


dae:
Most features in the FSRS crate are exposed via the methods of the FSRS struct, which takes and checks weights when it's initalized, so the checks can be a single time, at the start. Would it make sense to turn simulate() into another FSRS instance method instead?

L-M-Sherlock:
> Would it make sense to turn simulate() into another FSRS instance method instead?

I think it cannot solve the bug because the affected method belongs to `Card`.

L-M-Sherlock:
I find that we didn't verify the parameters if the user modified it manually. Maybe we need apply the `check_and_fill_parameters` once the user modify parameters?

dae:
simulate() appears to already call check_and_fill at the top. The problem is that create_review_priority_fn() clones a copy of the params prior to simulate being called. Instead of baking them into the function, why not have the function be passed a reference to the already-sanitized params?

L-M-Sherlock:
Seems we can read the parameters from `self.model.w` instead of passing parameters to the `simulator`?

dae:
Does that help here? The review priority function doesn't have access to the model either.

L-M-Sherlock:
I'm not sure whether I understand your solution. If we want the function be passed a reference to the already-sanitized params, we need read the params from FSRS instance (fsrs.model.w), right?

dae:
```diff
diff --git a/src/optimal_retention.rs b/src/optimal_retention.rs
index ff5223b..f9c2228 100644
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -67,7 +67,7 @@ impl std::fmt::Debug for PostSchedulingFn {
 /// Function type for review priority calculation that takes a card reference
 /// and returns a priority value (lower value means higher priority)
 #[derive(Clone)]
-pub struct ReviewPriorityFn(pub Arc<dyn Fn(&Card) -> i32 + Sync + Send>);
+pub struct ReviewPriorityFn(pub Arc<dyn Fn(&Card, &Parameters) -> i32 + Sync + Send>);
 
 impl PartialEq for ReviewPriorityFn {
     fn eq(&self, _: &Self) -> bool {
@@ -83,7 +83,7 @@ impl std::fmt::Debug for ReviewPriorityFn {
 
 impl Default for ReviewPriorityFn {
     fn default() -> Self {
-        Self(Arc::new(|card| (card.difficulty * 100.0) as i32))
+        Self(Arc::new(|card, _w| (card.difficulty * 100.0) as i32))
     }
 }
 
@@ -341,9 +341,10 @@ pub fn simulate(
     fn card_priority(
         card: &Card,
         learn: bool,
+        w: &Parameters,
         ReviewPriorityFn(cb): &ReviewPriorityFn,
     ) -> Reverse<(i32, bool, i32)> {
-        let priority = cb(card);
+        let priority = cb(card, w);
         // high priority for early due, review, custom priority
         Reverse((card.due as i32, learn, priority))
     }
@@ -354,6 +355,7 @@ pub fn simulate(
             card_priority(
                 card,
                 card.last_date == f32::NEG_INFINITY,
+                w,
                 &review_priority_fn,
             ),
         );
@@ -404,7 +406,7 @@ pub fn simulate(
             card.due = day_index as f32 + 1.0;
             card_priorities.change_priority(
                 &card_index,
-                card_priority(card, is_learn, &review_priority_fn),
+                card_priority(card, is_learn, w, &review_priority_fn),
             );
             continue;
         }
@@ -519,8 +521,10 @@ pub fn simulate(
             due_cnt_per_day[card.due as usize] += 1;
         }
 
-        card_priorities
-            .change_priority(&card_index, card_priority(card, false, &review_priority_fn));
+        card_priorities.change_priority(
+            &card_index,
+            card_priority(card, false, w, &review_priority_fn),
+        );
     }
 
     /*dbg!((
@@ -1559,37 +1563,49 @@ mod tests {
         run_test!(None, 42.339848)?;
         println!("High difficulty cards reviewed first.");
         run_test!(
-            wrap!(|card: &Card| -(card.difficulty * 100.0) as i32),
+            wrap!(|card: &Card, _w: &Parameters| -(card.difficulty * 100.0) as i32),
             43.859474
         )?;
         println!("Low retrievability cards reviewed first.");
         run_test!(
-            wrap!(|card: &Card| (card.retrievability(&DEFAULT_PARAMETERS) * 1000.0) as i32),
+            wrap!(|card: &Card, w: &Parameters| (card.retrievability(w) * 1000.0) as i32),
             43.482998
         )?;
         println!("High retrievability cards reviewed first.");
         run_test!(
-            wrap!(|card: &Card| -(card.retrievability(&DEFAULT_PARAMETERS) * 1000.0) as i32),
+            wrap!(|card: &Card, w: &Parameters| -(card.retrievability(w) * 1000.0) as i32),
             41.110588
         )?;
         println!("High stability cards reviewed first.");
         run_test!(
-            wrap!(|card: &Card| -(card.stability * 100.0) as i32),
+            wrap!(|card: &Card, _w: &Parameters| -(card.stability * 100.0) as i32),
             41.006256
         )?;
         println!("Low stability cards reviewed first.");
         run_test!(
-            wrap!(|card: &Card| (card.stability * 100.0) as i32),
+            wrap!(|card: &Card, _w: &Parameters| (card.stability * 100.0) as i32),
             43.077534
         )?;
         println!("Long interval cards reviewed first.");
-        run_test!(wrap!(|card: &Card| -card.interval as i32), 40.92563)?;
+        run_test!(
+            wrap!(|card: &Card, _w: &Parameters| -card.interval as i32),
+            40.92563
+        )?;
         println!("Short interval cards reviewed first.");
-        run_test!(wrap!(|card: &Card| card.interval as i32), 43.282585)?;
+        run_test!(
+            wrap!(|card: &Card, _w: &Parameters| card.interval as i32),
+            43.282585
+        )?;
         println!("Early scheduled due cards reviewed first.");
-        run_test!(wrap!(|card: &Card| card.scheduled_due() as i32), 43.309185)?;
+        run_test!(
+            wrap!(|card: &Card, _w: &Parameters| card.scheduled_due() as i32),
+            43.309185
+        )?;
         println!("Late scheduled due cards reviewed first.");
-        run_test!(wrap!(|card: &Card| -card.scheduled_due() as i32), 42.79738)?;
+        run_test!(
+            wrap!(|card: &Card, _w: &Parameters| -card.scheduled_due() as i32),
+            42.79738
+        )?;
         Ok(())
     }
 
```

dae:
```diff
diff --git a/rslib/src/scheduler/fsrs/simulator.rs b/rslib/src/scheduler/fsrs/simulator.rs
index d04e57125..144741f0a 100644
--- a/rslib/src/scheduler/fsrs/simulator.rs
+++ b/rslib/src/scheduler/fsrs/simulator.rs
@@ -75,7 +75,6 @@ pub(crate) fn apply_load_balance_and_easy_days(
 fn create_review_priority_fn(
     review_order: ReviewCardOrder,
     deck_size: usize,
-    params: Vec<f32>,
 ) -> Option<ReviewPriorityFn> {
     // Helper macro to wrap closure in ReviewPriorityFn
     macro_rules! wrap {
@@ -86,28 +85,28 @@ fn create_review_priority_fn(
 
     match review_order {
         // Ease-based ordering
-        EaseAscending => wrap!(|c| -(c.difficulty * 100.0) as i32),
-        EaseDescending => wrap!(|c| (c.difficulty * 100.0) as i32),
+        EaseAscending => wrap!(|c, _w| -(c.difficulty * 100.0) as i32),
+        EaseDescending => wrap!(|c, _w| (c.difficulty * 100.0) as i32),
 
         // Interval-based ordering
-        IntervalsAscending => wrap!(|c| c.interval as i32),
-        IntervalsDescending => wrap!(|c| -(c.interval as i32)),
+        IntervalsAscending => wrap!(|c, _w| c.interval as i32),
+        IntervalsDescending => wrap!(|c, _w| -(c.interval as i32)),
         // Retrievability-based ordering
         RetrievabilityAscending => {
-            wrap!(move |c| (c.retrievability(&params) * 1000.0) as i32)
+            wrap!(move |c, w| (c.retrievability(w) * 1000.0) as i32)
         }
         RetrievabilityDescending => {
-            wrap!(move |c| -(c.retrievability(&params) * 1000.0) as i32)
+            wrap!(move |c, w| -(c.retrievability(w) * 1000.0) as i32)
         }
 
         // Due date ordering
         Day | DayThenDeck | DeckThenDay => {
-            wrap!(|c| c.scheduled_due() as i32)
+            wrap!(|c, _w| c.scheduled_due() as i32)
         }
 
         // Random ordering
         Random => {
-            wrap!(move |_| rand::thread_rng().gen_range(0..deck_size) as i32)
+            wrap!(move |_c, _w| rand::thread_rng().gen_range(0..deck_size) as i32)
         }
 
         // Not implemented yet
@@ -197,7 +196,7 @@ impl Collection {
             .review_order
             .try_into()
             .ok()
-            .and_then(|order| create_review_priority_fn(order, deck_size, req.params.clone()));
+            .and_then(|order| create_review_priority_fn(order, deck_size));
 
         let config = SimulatorConfig {
             deck_size,

```

L-M-Sherlock:
> (it's an API breakage though, so will need a version bump)

So I need to bump the version to 4.0.0 if we adopt these changes?

dae:
Yep, since PostSchedulingFn has changed in the public API.

L-M-Sherlock:
continue on https://github.com/open-spaced-repetition/fsrs-rs/pull/323

PR #324, opened 2025-04-27
Feat/Make cards passable to CMRR

Luc-Mcgrady:
`v4.0.0` came faster than expected...

I doubt this will help with the `70%` issue but I still want to pr it while I have the opportunity.

### DIFF
--- a/benches/simulation.rs
+++ b/benches/simulation.rs
@@ -28,7 +28,7 @@ pub(crate) fn parallel_simulate(config: &SimulatorConfig) -> Result<Vec<f32>, FS
 }
 
 pub(crate) fn optimal_retention(inf: &FSRS, config: &SimulatorConfig) -> f32 {
-    inf.optimal_retention(config, &[], |_v| true).unwrap()
+    inf.optimal_retention(config, &[], |_v| true, None).unwrap()
 }
 
 pub fn criterion_benchmark(c: &mut Criterion) {
--- a/src/optimal_retention.rs
+++ b/src/optimal_retention.rs
@@ -1646,17 +1646,13 @@ mod tests {
             lapses: 0,
         };
         assert!((optimal_retention - 0.7).abs() < 0.01);
-        assert!(
-            fsrs.optimal_retention(&config, &[1.], |_| true, None)
-                .is_err()
-        );
+        fsrs.optimal_retention(&config, &[1.], |_| true, None)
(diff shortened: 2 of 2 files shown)


L-M-Sherlock:
> `v4.0.0` came faster than expected...

I haven't released it to crates.io, so it's OK.

dae:
This seems a bit messy at the moment. simulate() mutates the provided cards, and we invoke simulate() multiple times in parallel, leading to duplicate work.

Luc-Mcgrady:
> simulate() mutates the provided cards, and we invoke simulate() multiple times in parallel, leading to duplicate work.

I don't think the work is duplicated because cards have different desired retentions leading to different intervals? I think the workload here is the same since a new vector of cards has to be initialised with or without this change?

dae:
It is copying existing cards + creating new ones each time it is called. And it's called multiple times in parallel by sample().

Luc-Mcgrady:
https://github.com/open-spaced-repetition/fsrs-rs/blob/ed9567087dc87cae38376e14cf634467916d0b9c/src/optimal_retention.rs#L322-L334

But is copying the cards in such a way a problem compared to the way it's already done?

dae:
The inefficiency with existing cards is newly introduced by you. The inefficiency with created cards is not new, but I noticed it because of the former issue. The current approach is a bit messier than doing that work once up front , but I don't know whether it would have a noticeable difference on performance, and would require a bunch of refactoring. If Jarrett is happy with the code as-is and you don't notice a big difference when benchmarking, then I won't object.

L-M-Sherlock:
We don't have a benchmark for simulation with tons of existing cards.

The current benchmark only covered CMRR and simulation without existing cards.

https://github.com/open-spaced-repetition/fsrs-rs/blob/ed9567087dc87cae38376e14cf634467916d0b9c/benches/simulation.rs#L8-L28

https://github.com/open-spaced-repetition/fsrs-rs/blob/ed9567087dc87cae38376e14cf634467916d0b9c/benches/simulation.rs#L30-L32

dae:
On second thought, don't worry about the benchmark. Given that we were already repeating work for non-existing cards, it's unlikely to make a big difference.

L-M-Sherlock:
OK. Let me merge this PR.

L-M-Sherlock (approved):

PR #330, opened 2025-05-21
add unit test of `memory state`

ishiko732:
I wish I had a baseline memory state to use as a reference... 👉👈

ref:
- https://github.com/open-spaced-repetition/py-fsrs/pull/104#discussion_r2096092411
- https://github.com/open-spaced-repetition/ts-fsrs/pull/187


L-M-Sherlock (approved):
LGTM

PR #335, opened 2025-05-28
Feat/add termination probability parameter to expected_workload

L-M-Sherlock:


Expertium:
Is 0.01 sufficient?

L-M-Sherlock:
> Is 0.01 sufficient?

It's configurable. You can pass different value in Anki's side.

Issue #334, opened 2025-05-27
Publish 4.0.0 to crate.io?


L-M-Sherlock:
3.0.0 was published one month ago. If we don't have anymore breaking changes, I plan to publish 4.0.0 to crate.io. I hope it could follow up the next release of Anki.

@Luc-Mcgrady @dae any thoughts?

PR #342, opened 2025-06-11
External cmrr target

Luc-Mcgrady:
Lets you specify a different target for CMRR.
Ticks the bottom box of https://github.com/open-spaced-repetition/fsrs-rs/issues/334#issuecomment-2914518036

@asukaminato0721 would you mind reviewing this?


src/optimal_retention.rs
```diff
@@ -682,6 +708,7 @@ fn sample<F>(
     n: usize,
     progress: &mut F,
     cards: &Option<Vec<Card>>,
+    target: &CMRRTargetFunction,
```
asukaminato0721, line 711:
```suggestion
    CMRRTargetFunction(target): &CMRRTargetFunction,
```

I guess we can try like this.

asukaminato0721:
ref

https://github.com/open-spaced-repetition/fsrs-rs/blob/6f3f5707ce620c12504916d9c92f950b933e255e/src/optimal_retention.rs#L463

and `Function` or `Fn`? I want to unify them :)

L-M-Sherlock (approved):
LGTM

PR #352, opened 2025-07-22
Refactor/expected workload via dp

L-M-Sherlock:
refer to: https://github.com/open-spaced-repetition/expected-workload/blob/caa3ff96092be8b913005c358ba023bc6feb9e7a/DynamicProgramming.ipynb


L-M-Sherlock:
Great! Could you test it on the user who reported the workload problem before?

PR #354, opened 2025-07-31
Fix/clamp the minimum of w[19]

L-M-Sherlock:
same to https://github.com/open-spaced-repetition/fsrs-optimizer/pull/186


src/parameter_clipper.rs
```diff
@@ -38,6 +47,10 @@ pub(crate) fn clip_parameters(parameters: &Parameters, num_relearning_steps: usi
     } else {
         2.0
     };
+    let w19_floor = match enable_short_term {
+        Some(true) => 0.01,
+        Some(false) | None => 0.0,
```
asukaminato0721, line 52: I don't know why there need 3 status: Some true, Some false, None.

asukaminato0721 (approved):

PR #367, opened 2025-09-29
Refactor/extract standalone functions from structure

L-M-Sherlock:


Luc-Mcgrady (commented):
Should `evaluate_with_time_series_splits` and `optimal_retention` be made standalone? 

also should we remove the `Option` in

https://github.com/open-spaced-repetition/fsrs-rs/blob/1e271981367454468391f1c686af03a0aa7aab3c/src/model.rs#L256

if the intention is that there are no functions that don't require parameters?

asukaminato0721 (approved):

Issue #391, opened 2026-02-22
Make it possible to resume optimization


Expertium:
<img width="558" height="69" alt="Image" src="https://github.com/user-attachments/assets/f6965f0f-b606-41d2-aa97-4b6df4c7786f" />

This got me thinking, and after discussing it with LLMs, here's my takeaway. If we could do ALL of this:

1) Store Adam's state

2) Store `best_loss` and `best_model` for every epoch

3) Make `ShuffleDataLoader` deterministic

4) Make `CosineAnnealingLR` independent of the dataset size somehow

Then this would be doable.
1 and 2 are trivial. 4 isn't straightforwardly doable because cosine annealing requires `t_max`, but we could replace cosine annealing with exponential decay, which would probably hurt log loss a little bit, but would eliminate reliance on `t_max`.
That leaves 3, and I have no idea how hard that is.
I think this is worth pursuing. If we don't have to do optimization from zero every time and instead only need to run optimization on new reviews, this could dramatically speed optimization up.

EDIT: I completely forgot about the recency weighting.

5) Make recency weighting independent of the dataset size somehow 

No idea how this could possibly be done. It's literally defined in terms of first review - last review, so we're kinda screwed.

EDIT 2: We could do something like this:

w = w_min⁡ + (1 - w_min⁡)⋅(1-e^(-λ⋅t)),

where t is the number of days since the oldest review, which gets t=0. So the oldest review has t=0, the review 10 days after the oldest review gets t=10, the review one year after the oldest review gets t=365. Now adding a new review doesn't change the weights assigned to all other reviews.

<img width="1402" height="814" alt="Image" src="https://github.com/user-attachments/assets/47ea6733-9785-4ffe-b119-f8f550e8778a" />

The shape you get as the result is pretty different from what is optimal for recency weighting though. And if λ is chosen to work well on average, for users with very old collections most weights will be very close to 1, defeating the purpose of recency weighting.

EDIT 3: I guess we can kinda monkey patch it with another parameter. But it's ugly as sin.

w = w_min⁡ + (1 - w_min⁡)⋅(1-e^(-λ⋅max(t - t_max, 0)))

<img width="1402" height="814" alt="Image" src="https://github.com/user-attachments/assets/6c165e65-a8ad-46f3-b63c-737f0121e2ee" />

[...]

PR #394, opened 2026-03-27
Make it possible to resume optimization

MoonMao42:
Attempt at #391.

This makes the current training flow resumable. It stores the model, best model, optimizer, scheduler, and shuffle state in a checkpoint, and adds `compute_parameters_resumable()` / `resume_compute_parameters()` so interrupted runs can continue later.

If you'd prefer a different direction here, happy to discuss.

### DIFF
--- a/src/batch_shuffle.rs
+++ b/src/batch_shuffle.rs
@@ -6,6 +6,7 @@ use burn::prelude::Backend;
 use rand::SeedableRng;
 use rand::rngs::StdRng;
 use rand::seq::SliceRandom;
+use serde::{Deserialize, Serialize};
 
 use crate::dataset::{FSRSBatch, FSRSBatcher, FSRSDataset};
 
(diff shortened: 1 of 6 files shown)


asukaminato0721:
Thanks for the PR. I reviewed it against issue #391, and my read is that it solves the narrower problem of resuming an interrupted optimization run from a checkpoint, but it does not fully satisfy the broader goal described in #391.

What this PR does well:
  - It adds resumable optimization checkpoints that capture the model, optimizer, best model, LR scheduler, shuffle state, and frozen training data.
  - It restores those states and continues training from the interruption point.
  - The tests show that resuming after an interrupt can match a fresh uninterrupted run.

What it does not yet solve from #391:
  - It does not support resuming optimization with newly added reviews. resume_compute_parameters resumes only from the frozen dataset stored in the checkpoint, so this is not incremental optimization over updated data.
  - The training flow still relies on dataset-size-dependent cosine annealing and dataset-size-dependent recency weighting, which were two of the main blockers called out in #391.

So I would describe this PR as an important building block toward #391, but not a complete resolution of that issue as originally framed. If we merge it, I think the scope should be described as “resume interrupted optimization from checkpoint,” with the broader “incremental optimization on new reviews” work left for follow-up.

There are also a few correctness issues to fix before merge:
  - Progress is not always marked finished on error paths, which can leave progress waiters hanging.
  - Resumed progress has the same problem on checkpoint/load errors.
  - Checkpoint shuffle-state validation should reject impossible states instead of silently accepting malformed data.

@Expertium, what's your opinion here?

Expertium:
I closed that issue as not planned.

PR #395, opened 2026-04-07
FSRS-7 Support

JSchoreels:
Ongoing PR with already quite a bit of discussion/work made with @Expertium and Alex on discord.

Basically right now there's a penalty and one without penalty to avoid 1s hell review.

I've integrated this version in Anki and it works fine, but the fact S is not the interval at R=90% make it a bit messy to handle in Anki, but we can focus on this later. I added util method to compute a S90 which is the "legacy S"

There's some caching implementing with a look up table to avoid having to redo each time the S->S90


Expertium:
Btw, check out some of the recent commits I made. FSRS-rs in the benchmark needs to support this flag too
https://github.com/open-spaced-repetition/srs-benchmark/commit/f843e7a0ddbca744297f5a7dedd8fdd4ebab7851

asukaminato0721:
I reviewed the patch and found two behavioral regressions that should be fixed before merge.

1. FSRS::default() / FSRS::new(&[]) now instantiate a 35-parameter FSRS-7 model, but memory_state_from_sm2() still reads the legacy FSRS-6 slots. That breaks the existing SM-2 conversion path for callers relying on the default constructor and changes runtime behavior in a backwards-incompatible way.
2. enable_short_term=false is no longer honored for 35-parameter training runs. The FSRS-7 clipping path ignores that flag, and the FSRS-7 transition logic still blends the short-term block into step(), so the opt-out no longer produces a purely long-term model.

I’d also strongly recommend splitting model.rs as part of this work. Right now FSRS-6 and FSRS-7 logic live in the same file, which makes the code harder to reason about and increases the chance of compatibility regressions like the ones above. Separating the two implementations into distinct modules would make the branching clearer, reduce file size, and make it much easier to verify that legacy FSRS-6 behavior remains intact while FSRS-7 evolves.

The main issue is not the move toward FSRS-7 itself, but that the new default behavior was switched over before the legacy compatibility paths were fully migrated. The safest fix would be to preserve the old default constructor behavior until the SM-2 bridge and the short-term opt-out semantics are fully ported.

src/simulation.rs
```diff
@@ -204,30 +204,91 @@ fn init_s(w: &[f32], rating: usize) -> f32 {
...
-    let hard_penalty = if rating == 2 { w[15] } else { 1.0 };
-    let easy_bonus = if rating == 4 { w[16] } else { 1.0 };
-    (s * (f32::exp(w[8])
-        * (11.0 - d)
-        * s.powf(-w[9])
-        * (f32::exp((1.0 - r) * w[10]) - 1.0)
-        * hard_penalty)
-        .mul_add(easy_bonus, 1.0))
-    .clamp(S_MIN, S_MAX)
+fn is_fsrs7(w: &[f32]) -> bool {
```
Luc-Mcgrady: We could use a `SimulatedFsrs` trait for this with `LegacySimulatedFsrs` + `SimulatedFsrs7` implementations rather than checking the length every function call.

something like

```rs
trait SimulatedFsrs {
    fn test(&self) -> &'static str;
}

struct SimulatedFsrs7;
struct LegacySimulatedFsrs;

impl SimulatedFsrs for SimulatedFsrs7 {
    fn test(&self) -> &'static str {
        "FSRS 7"
    }
}

impl SimulatedFsrs for LegacySimulatedFsrs {
    fn test(&self) -> &'static str {
        "Legacy FSRS"
    }
}

fn main() {
    let is_fsrs_7 = true;
    let fsrs: &dyn SimulatedFsrs = if is_fsrs_7 { &SimulatedFsrs7 } else { &LegacySimulatedFsrs };
    println!("{}", fsrs.test());
}

```

JSchoreels:
@asukaminato0721 : Thank you for the feedback, I've reverted the default to FSRS6 and added a memory_state_from_sm2 for fsrs7. I've also done the split between model v6/v7 as best as I could, and tried to do the similar work for other files.

@Luc-Mcgrady I tried to reduce the amount of parameter length checks to know if it's FSRS6/7, and used traits.

@Expertium , I've integrated the flag and asked codex to check to adapt with new changes from srs-benchmark

src/simulation.rs
Luc-Mcgrady: You're still doing an if statement and checking the length for every function call. My idea was there should be something analgous to "let fsrs: &dyn model = model_v6 or model_v7" in the main function.

JSchoreels: alright, as discussed I've done it but for  the simulator only since with Burn it's not that easy to do.

JSchoreels:
@Luc-Mcgrady : Implemented your suggestion but for simulator only, the Burn library make it difficult to do the same for the optimization

@Expertium : Implemented the fractional interval for fsrs7. To make it easier in terms of datatype, I've used float for everything, fsrs6 will simply round the values to make interval (like in anki).

I duplicated test that were modified from 6 to 7 to make them run for 6 AND 7, to be sure both fsrs still work as intended

JSchoreels:
For reference, on my own collection 

FSRS6 params:
> 0.0535, 1.1503, 2.9353, 9.7803, 6.6401, 0.4367, 2.2582, 0.0018, 1.5340, 0.2589, 0.5450, 1.3762, 0.1718, 0.4944, 0.8939, 0.3181, 5.0850, 0.7925, 0.3163, 0.1312, 0.2103

FSRS7 params : 
> 0.1146, 1.8036, 4.1998, 18.6138, 5.7531, 0.4934, 2.4162, 1.9752, 0.2789, 1.0798, 0.7342, 0.0096, 0.7866, 0.4458, 0.3313, 3.1915, 1.0442, 0.6311, 3.7169, 0.8447, 0.0017, 0.7010, 2.6546, 0.5310, 1.3865, 2.5147, 0.9910, 0.0559, 0.3496, 0.6166, 0.9225, 0.1879, 0.6955, 0.0280, 0.1867

CleanShot 2026-04-09 at 10 39 59@2x

I have already branches of Anki, FSRS Helper, ts-fsrs, Search Stats Extended with all FSRS7 selectable, since I added a FSRS version selector in my anki fork

image

To run this, please check all my forks (branch main for all)

Impacted Projects:
- [jschoreels/fsrs-rs](https://github.com/JSchoreels/fsrs-rs)
  - Integrate srs-benchmark models/fsrs_v7.py
- [jschoreels/anki](https://github.com/JSchoreels/anki)
  - Plugged-in deck options, simulator, help me decide, and how Retrievability is retrieved from DB (previously used a lighter computation based on a single decay curve; FSRS7 uses multiple)
  - Change Deck Options to include an FSRS selector (older versions may still be bugged)
- [jschoreels/fsrs4anki-helper](https://github.com/jschoreels/fsrs4anki-helper)
- [jschoreels/ts-fsrs](https://github.com/JSchoreels/ts-fsrs)
- [jschoreels/Anki-Search-Stats-Extended](https://github.com/jschoreels/Anki-Search-Stats-Extended)

src/lib.rs
```diff
@@ -18,8 +19,8 @@ mod training;
 pub use dataset::{FSRSItem, FSRSReview, filter_outlier};
 pub use error::{FSRSError, Result};
 pub use inference::{
-    DEFAULT_PARAMETERS, FSRS5_DEFAULT_DECAY, FSRS6_DEFAULT_DECAY, ItemProgress, ItemState,
-    MemoryState, ModelEvaluation, NextStates, current_retrievability,
+    DEFAULT_PARAMETERS, FSRS5_DEFAULT_DECAY, FSRS6_DEFAULT_DECAY, FSRS6_DEFAULT_PARAMETERS,
+    ItemProgress, ItemState, MemoryState, ModelEvaluation, NextStates, current_retrievability,
```
ishiko732, line 23: Do we need to support both FSRS6 and FSRS7 within the same version?

JSchoreels: Personally, I think it makes sense. FSRS7 is much more complex in how the decay works and can have a dramatic effect on very high DR. An app developer could decide to handle both versions. FSRS6 under the hood was kind of supporting previous FSRS too, for example, in Anki, for fallback with older clients.

FSRS7 "breaks" certain things like Stability being the duration at which Retrievability drop from 100% to 90%, so it's also probably safer to have both versions able to run in parallel, in my opinion.

So I think it makes sense to be able to support both.

JSchoreels:
There was a regression in FSRS-6 ... The regularization was off apparently, and then I started doing parity checks (before and after this PR) and it seemed something changed too, it was apprently some smoothing function that got lost during the refactor..

This is the results of the tests before/after for the 21 params

| Param | Before upstream | After current | Abs diff | Rel diff | Most likely explanation |
|---:|---:|---:|---:|---:|---|
| 0 | 0.0010 | 0.0001 | 0.0009 | 90.00% | `S_MIN` changed. |
| 1 | 0.0010 | 0.0001 | 0.0009 | 90.00% | `S_MIN` changed. |
| 2 | 0.0010 | 0.0001 | 0.0009 | 90.00% | `S_MIN` changed. |
| 3 | 0.0010 | 0.0001 | 0.0009 | 90.00% | `S_MIN` changed; previous init/smoothing drift is fixed. |
| 4 | 6.50 | 6.46 | 0.04 | 0.59% | Optimizer/config drift. |
| 5 | 0.73 | 0.74 | 0.01 | 0.88% | Optimizer/config drift. |
| 6 | 3.29 | 3.18 | 0.11 | 3.23% | Optimizer/config drift. |
| 7 | 0.0010 | 0.0010 | 0.00 | 0.00% | Matches. |
| 8 | 1.72 | 1.77 | 0.06 | 3.28% | Optimizer/config drift. |
| 9 | 0.07 | 0.10 | 0.03 | 41.79% | Small baseline amplifies relative difference; optimizer/config drift. |
| 10 | 0.64 | 0.70 | 0.06 | 8.98% | Optimizer/config drift. |
| 11 | 1.33 | 1.39 | 0.05 | 3.98% | Optimizer/config drift. |
| 12 | 0.18 | 0.15 | 0.03 | 18.42% | Small baseline amplifies relative difference; optimizer/config drift. |
| 13 | 0.10 | 0.16 | 0.06 | 59.65% | Small baseline amplifies relative difference; optimizer/config drift. |
| 14 | 1.52 | 1.57 | 0.05 | 3.14% | Optimizer/config drift. |
| 15 | 0.81 | 0.72 | 0.09 | 11.02% | Optimizer/config drift. |
| 16 | 2.14 | 2.01 | 0.13 | 6.28% | Largest residual; optimizer/config drift. |
| 17 | 0.55 | 0.53 | 0.01 | 2.56% | Optimizer/config drift. |
| 18 | 0.08 | 0.13 | 0.05 | 58.02% | Small baseline amplifies relative difference; optimizer/config drift. |
| 19 | 0.07 | 0.07 | 0.00 | 0.00% | Matches. |
| 20 | 0.10 | 0.10 | 0.00 | 0.00% | Matches. |

JSchoreels:
FYI, I have a running anki fork with this fsrs-rs targeted, if you want to try it out : https://github.com/JSchoreels/anki/releases

Expertium:
@JSchoreels @L-M-Sherlock should we merge this first, or wait for me to finish enhancing FSRS-7 formulas (about 3 more days)?

PR #417, opened 2026-06-04
perf: speed up compute_parameters (windowed analytic gradient)

Expertium:
Written by Claude.

Three bit-for-bit optimizations to the card-grouped (`card_ids`) analytic
gradient path that dominates `compute_parameters`, building on the analytic
gradient from #411.

### Changes
1. **Hoist loop-invariant transcendentals** out of the per-step recurrence
   (curve factor `exp(ln0.9/decay)`, `exp(w8)`, `exp(w17*w18)`,
   `init_difficulty(w, 4)`) — computed once per call instead of once per
   (card x timestep).
2. **Evaluate only the active stability branch** in the forward step. It used
   to compute all three (success / lapse / same-day) and select one; since the
   selection and `step_backward` dispatch on the same flags, the unused
   branches were wasted transcendentals.
3. **Skip trailing-padding timesteps.** Cards are length-sorted and batched, so
   shorter cards carry trailing padding (rating 0) up to the batch-wide
   `seq_len`. That padding is a zero-gradient passthrough, yet the recurrence
   still evaluated a forgetting curve at each padding step — so stop each column
   at its real length.

### Correctness
All three are bit-for-bit: trained parameters and log loss are unchanged. The
`finite_difference_matches_batch_gradient`, `windowed_matches_sum_of_prefixes`,
and `test_analytic_loss_and_grad_matches_burn` tests pass (rebased on top of
\#416).

### Benchmark
`cargo bench --bench parameters` on a ~600k-review collection:
- `compute_parameters` (windowed): **about -22%**
- `compute_parameters_without_card_ids`: **about -30%**

### One more (~7%) I'd like your call on
The gradient path computes a BCE loss value that `train()` discards (only the
adjoint feeds the backward). Skipping it saves two `ln` per scored step
(about 7% on the bench here). But `test_analytic_loss_and_grad_matches_burn` and
`windowed_matches_sum_of_prefixes` assert that returned loss, so it can't be
dropped without either a `const SCORE_LOSS: bool` split (training skips the
loss; tests/validation keep it) or having those tests assert only the gradient.
Would you be open to either? Happy to send a follow-up.

### DIFF
--- a/src/analytic.rs
+++ b/src/analytic.rs
@@ -715,6 +715,13 @@ fn card_column_loss_and_grad(
     };
     for t in 0..seq_len {
         let idx = t * batch + column;
+        // Trailing padding (rating 0) is a no-op passthrough: it scores nothing (weight 0) and
+        // carries a zero adjoint back through the state clamp, so it contributes neither loss nor
+        // gradient. A column's reviews are contiguous and then padded, so once the first padding
+        // step is hit the rest are all padding — stop at the card's real length. Bit-for-bit.
+        if r_hist[idx] == 0.0 {
+            break;
+        }
         let (next, cache) = step_forward(w, consts, state, t_hist[idx], r_hist[idx], t);
         let (step_loss, g_r) = if t == 0 {
             (0.0, 0.0)
--- a/src/analytic.rs
+++ b/src/analytic.rs
@@ -215,43 +215,58 @@ fn step_forward(
     let curve = curve_forward(w, delta_t, last_s, consts.factor);
     let r = curve.retrievability;
 
-    let hard_penalty = if rating == 2.0 { w[15] } else { 1.0 };
-    let easy_bonus = if rating == 4.0 { w[16] } else { 1.0 };
-    let success_inc = consts.exp_w8
-        * (11.0 - last_d)
-        * last_s.powf(-w[9])
-        * (((1.0 - r) * w[10]).exp() - 1.0)
-        * hard_penalty
-        * easy_bonus;
-    let success = last_s * (success_inc + 1.0);
-
-    let failure_raw = w[11]
-        * last_d.powf(-w[12])
-        * ((last_s + 1.0).powf(w[13]) - 1.0)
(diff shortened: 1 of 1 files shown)


Expertium:
This was made by Claude on my behalf

Expertium:
Written by Claude.

One more optimization I measured, in addition to the dead-loss skip noted in
the description:

**f32 backward pass (~4-5%).** The forward is already f32, but the backward
computes the VJP in f64. Doing it in f32 -- while keeping the `gw` accumulator
in f64 so the cross-card sum stays precise -- is ~4-5% faster on the windowed
`compute_parameters`, with a mean log-loss change of **+1e-6** (negligible; the
model trains in f32 anyway, and only 5/50 parameters move at all). It passes
`test_analytic_loss_and_grad_matches_burn` (2e-3 tolerance) and the
finite-difference test, but fails `windowed_matches_sum_of_prefixes`: that test
asserts the O(N) and O(N^2) gradients match within **1e-9**, and f32 rounding
makes them diverge by ~3e-9. So it would need that tolerance loosened to an
f32-appropriate value.

Like the loss-skip, I left it out of this PR since it touches a test guarantee
-- just flagging it in case you would rather take the ~5% and relax that bound.
Happy to send a follow-up either way.

Expertium:
Written by Claude.

Pushed one more bit-for-bit commit: **skip the forgetting curve for same-day
reviews**. For `delta_t == 0` the curve base is exactly 1, so retrievability is
1 regardless of the parameters -- `1.powf(x)` is always 1, so that powf is pure
waste. These steps are common (a card's first step, plus every same-day learning
step), so it's a real win on collections with many same-day reviews (~+10% on a
same-day-heavy dataset; less on collections that are mostly mature spaced
reviews). All gradient tests pass unchanged, including the windowed-vs-prefix
1e-9 exact-match -- `1.powf` is exact, so nothing moves.

L-M-Sherlock:
duplicated of #419

PR #419, opened 2026-06-05
Speed up compute_parameters without SIMD

asukaminato0721:
## Summary

This PR speeds up compute_parameters using scalar, architecture-independent optimizations only. It does not add SIMD, NEON, wide, or target-specific code.

Changes include:
- Reuse train batches for validation when compute_parameters trains and validates on the same dataset.
- Avoid redundant full-dataset cloning and card-id realignment during outlier filtering.
- Hoist scalar runtime constants in the analytic training path.
- Skip dead padding/init work and inactive stability branches in step_forward.
- Reuse scratch buffers during analytic gradient computation.
- Track real column lengths in host batches to skip trailing padding timesteps.
- Clip parameters in place inside the training loop.

## Performance

On the local Apple Silicon benchmark:

- compute_parameters: 27.756 ms -> 17.103 ms
- compute_parameters_without_card_ids: 79.110 ms -> 53.394 ms

Best observed run with the same retained code:

- compute_parameters: 16.036 ms
- compute_parameters_without_card_ids: 47.157 ms

## Verification

- cargo test --quiet
- cargo bench --bench parameters -- compute_parameters
- git diff --check
- Confirmed no SIMD/NEON/wide references were added

### DIFF
--- a/src/analytic.rs
+++ b/src/analytic.rs
@@ -43,6 +43,44 @@ struct State {
     d: f32,
 }
 
+#[derive(Clone, Copy)]
+struct RuntimeParams<'a> {
+    w: &'a [f32],
+    curve_decay: f32,
+    curve_factor: f32,
+    curve_dfactor_ddecay: f64,
+    exp_w8: f32,
+    exp_w8_f64: f64,
+    failure_floor_divisor: f32,
+    easy_d: f32,
+    easy_d_f64: f64,
+    exp_3w5_f64: f64,
+}
+
+impl<'a> RuntimeParams<'a> {
+    #[inline]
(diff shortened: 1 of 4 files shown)


asukaminato0721:
removes the unused validation-set path from training.

It changes train(...) to only accept the weighted training set, then reuses the already-built train_batches for epoch loss calculation. This removes the test_set: Option<_> parameter, the duplicate validation batch construction, and the extra weighted_train_set.clone() in benchmark(...).

Net effect: same current behavior, simpler code, less allocation/work in the benchmark path. Diff: src/training.rs, 3 insertions, 16 deletions.

L-M-Sherlock (approved):
LGTM

PR #426, opened 2026-06-14
FSRS7 + ADR + Optimization

JSchoreels:
This build on top of #395 and #412.

It mainly use @Expertium performance points, with a specific implementation of NEON for macos, and also improve by 15-20% ADR time computation.

PR #447, opened 2026-09-09
Remove Burn dependency after FSRS-7 optimization

JSchoreels:
Hello @asukaminato0721, I made this PR to followup on your request to remove Burn from the recent PR you merged from me.

## Summary

Follow-up to #426.

- Remove Burn from the normal and development dependency graphs.
- Replace the remaining tensor/autodiff inference and FSRS-6/FSRS-7 training paths with scalar analytic Rust implementations and host Adam.
- Retire the Burn-only neural revlog classifier target while retaining the host-only classifier example.
- Replace the generic FSRS backend API with the concrete FSRS::new constructor.
- Reduce the default Cargo dependency graph from 340 to 110 entries.

## Collection benchmark

Measured before and after with the same copied recent Anki collection backup and a 10-sample Criterion protocol:

| Training input | Before | After | Change |
| --- | ---: | ---: | ---: |
| Aligned card IDs | 213.25 ms | 219.04 ms | 2.7% slower |
| No card IDs | 80.486 s | 1.7183 s | 46.8x faster, 97.9% less time |

The aligned-card-ID path is the path used by normal Anki optimization, so this change primarily improves callers that do not provide card IDs.

## Validation

- cargo test --release --all-targets: 118 tests passed across 7 suites
- cargo check --all-targets
- cargo clippy -- -Dwarnings
- cargo fmt --all -- --check
- rustdoc with warnings denied
- all-features Cargo metadata audit: no Burn package
- git diff --check


L-M-Sherlock:
We need pass the tests at first.

